PR #7, opened 2015-05-07
make closingReason a property

morethanreal:
@cdata PTAL (behavior overriding stuff)

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -58,6 +58,7 @@ Fired after the `iron-overlay` closes.
        * True if the overlay was cancelled.
        */
       canceled: {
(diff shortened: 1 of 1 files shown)


cdata:
Can you briefly explain where the overriding problem was happening?

morethanreal:
`iron-overlay-behavior` fires an event with why the dialog is closed with the close event. Previously, the `closingReason` was a getter. `paper-dialog-behavior` adds an additional reason `confirmed` for whether somebody closed a dialog due to a dismissive or affirmative action (spec terms), and it implemented it by overriding the getter. With 0.9, that no longer works, so I switched it to a property.

cdata:
LGTM :+1:

PR #8, opened 2015-05-14
don't call _openedChanged if initially closed

morethanreal:
@cdata PTAL cc @notwaldorf 

Prevents doing extra work when the overlay starts closed.


cdata:
LGTM :ok_hand:

PR #17, opened 2015-06-09
fix event param doc

morethanreal:
For https://github.com/PolymerElements/paper-dialog-behavior/issues/11

@garlicnation @jklein24 PTAL

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -145,8 +145,8 @@ context. You should place this element as a child of `<body>` whenever possible.
 
 /**
  * Fired after the `iron-overlay` closes.
- * @event iron-overlay-closed {{canceled: boolean}} detail -
- *     canceled: True if the overlay was canceled.
+ * @event iron-overlay-closed
+ * @param {{canceled: (boolean|undefined)}} set to the `closingReason` attribute
  */
 
     listeners: {


jklein24:
Is switching the type to {{canceled: (boolean|undefined)}} OK? What you have here works fine and the compiler doesn't even look at @event annotations, but it's nice to have more specific does if possible.

morethanreal:
Yeah, that would be ok. I didn't write it this way in this PR because I have another behavior that includes this behavior that adds more keys to `closingReason`. Would I re-declare the event docs in the other behavior?

jklein24:
I guess that's up to @garlicnation and the hydrolysis folks. JSCompiler doesn't care about event annotations either way as long as there aren't invalid `@param` annotations.

garlicnation:
@jklein24 What's the recommended way to subclass in jscompiler? It seems like we'd want an "extends" or similar syntax for this use case.

jklein24:
`@extends` is there, but I'm not sure that will really work correctly with the compiler for event types... Where are you thinking that annotation would go?

garlicnation:
I was thinking the event type should get pulled out into a class or typedef, and referred to in this @param annotation, rather than being explicitly spelled out.

morethanreal:
@garlicnation What's the syntax to do that?

jklein24:
That sounds fine, but I'm still not sure where `@extends` would be useful

morethanreal:
@jklein24 @garlicnation What should be my next steps for this PR?

jklein24:
Either:

```
@event iron-overlay-closed
@param {{canceled: (boolean|undefined)}} detail ...
```

or extract the detail here into a typedef like:

```
/**
 * The detail object type for an iron-overlay-closed event.
 * canceled: True if the overlay was canceled.
 * @typedef {{canceled: (boolean|undefined)}}
 */
Polymer.IronOverlayBehaviorImpl.ClosedEventDetail;
```

PR #26, opened 2015-07-17
use event.path to determine whether an overlay should close

morethanreal:
PTAL @cdata


cdata:
@morethanreal Thanks for the PR. I think that this has been resolved by @azakus in another PR. Please re-open if you think it's still a problem.

Issue #27, opened 2015-07-22
closingReason.canceled not set correctly after closing overlay 2nd time


thansber:
If you open an overlay, then close it by clicking outside, the `closingReason.canceled` flag is set to true, all good so far.  If you open the overlay again, then close it by clicking outside, the `closingReason.canceled` flag is set to false, and any listeners to iron-overlay-closed receive `event.detail.canceled === false`.

I think the problem is the cancel method in `iron-overlay-behavior.html` is setting opened to false before setting canceled to true.  When I switched the lines within cancel, closingReason.canceled was set correctly and iron-overlay-closed listeners receive the proper canceled flag.

There is also a comma ending the first line in cancel, should probably be a semi-colon.  Also wondering if this change should be made to the close method as well.

Lines in question: https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L213-L214
JSBin: http://jsbin.com/refogupana/1/edit?html,output

thansber:
https://github.com/PolymerElements/iron-overlay-behavior/pull/28

thansber:
This issue is still occurring.

thansber:
The pull request no longer fixes the issue and I have been unable to fix it locally without breaking a lot of other specs.  I do have a spec that fails when simulating the behavior, should I create a pull request with this failing spec?

PR #28, opened 2015-07-22
set canceled before setting opened flag

thansber:
Opening an overlay a 2nd time and closing it causes the closingReason.canceled flag to be set to false when it should be true.

https://github.com/PolymerElements/iron-overlay-behavior/issues/27


thansber:
I signed it!

valdrinkoshi:
@thansber looks good, could you add unit tests to avoid future regressions?

thansber:
Sure thang.

thansber:
I messed up this pull request and will start a new one with the fix + tests.

valdrinkoshi:
sounds good, thanks!

PR #29, opened 2015-07-31
Suppress compiler warnings about suspicious code

rictic:
We have a good reason for doing what would otherwise be a noop accessor here, this just tells the compiler that.


jklein24:
Does this actually fix the errors? I thought suppressions need to be on function blocks.

rictic:
It does! You can suppress individual lines: https://github.com/google/closure-compiler/wiki/Warnings#suppress-tags

jklein24:
Hmm it seems some suppressions work work on single lines and others don't... interesting.

LGTM

PR #32, opened 2015-08-15
Update iron-overlay-behavior.html

aronfeerick:
Added "this.resetFit();" to "_onIronResize" function, I made this change on my projects and it fixes the problem where "paper-dialog" doesn't stay centralised when the window is resized.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -421,6 +421,7 @@ context. You should place this element as a child of `<body>` whenever possible.
         return;
       }
       if (this.opened) {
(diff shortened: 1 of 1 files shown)


aronfeerick:
I signed it!

aronfeerick:
On further investigation this fix only fixes the issues for the first "paper-dialog".

So to fix this I then had to add the following to my project.

$(window).on("resize",function(){
 $("paper-dialog").each(function(k){
  if(this.opened){this.resetFit();this.refit();}
 });
});

PR #34, opened 2015-08-20
Move @event declarations to bottom of prototype

rictic:
This is so that the compiler doesn't attribute the @param annotation to be referencing the next line (listeners in this case)


cdata:
Should we be doing this across the board for elements?

LGTM :+1:

rictic:
Maybe? I think @garlicnation has a proposal to declare events in a different way, so that this won't happen.

PR #35, opened 2015-09-02
Fix #21, iOS backdrop does not close

### LINKED ISSUE #21: Tapping on backdrop doesn't close overlay
`no-cancel-on-outside-click` has no apparent effect on mobile Safari. I was unable to close overlay by tapping outside, in the backdrop area.

atotic:
See #21 on iOS Safari, overlay does not close when clicking on backdrop
Cause:
- iron-overlay closes by listening to `click` on document.
- on iOS, `click` does not bubble, and never gets triggered (see http://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html)

My quick fix is to add a `touchstart` listener. I've tested it, and it works for me.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -269,6 +269,7 @@ context. You should place this element as a child of `<body>` whenever possible.
       }
       // async so we don't auto-close immediately via a click.
       this._toggleListenersAsync = this.async(function() {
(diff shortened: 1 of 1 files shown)


atotic:
use-tap pull request is a better fix than mine.

PR #36, opened 2015-09-02
Use a global tap handler, not click handler

azakus:
Fixes PolymerElements/paper-icon-button#35


atotic:
This also fixes #21, tested it. Gotta learn those Gestures. Still not sure how the fix works, since your `tap` event handler binds to _nop.

azakus:
Address the comments from the previous commit. The "_nop" was just to there to give addEventListener a function in `Polymer.Gestures.add`, which I can actually avoid.

The real handler is the capturing "tap" here: https://github.com/PolymerElements/iron-overlay-behavior/pull/36/files#diff-5e1626c54b6ff0c467b99e2307ad8af4R280

Issue #37, opened 2015-09-24
Event to know when the dialog is canceled


ebidel:
I'd like to distinguish between  a close and a cancel (from clicking outside the region).
Close happens to late, and I need to take action based on a cancel.

Firing `iron-overlay-canceled` would be ideal or even making `canceled` notify.

PR #41, opened 2015-10-02
Adds 'iron-overlay-canceled' event, calling `preventDefault` prevents…

bicknellr2:
… the overlay from closing. (Fix for #37.)


cdata:
This change looks great :) thanks for the contribution!

ebidel:
We need to document this event.

bicknellr2:
Ok, here's the PR: #42

PR #51, opened 2015-11-05
Add IronOverlayManager.ensureMinimumZ()

justinfagnani:


justinfagnani:
If this approach looks good, I can add tests.

iron-overlay-manager.html
```diff
@@ -77,35 +91,25 @@
...
-      trackBackdrop: trackBackdrop,
-      getBackdrops: getBackdrops
-    };
-
-  })();
+    },
+
+    getBackdrops: function() {
+      return this._backdrops;
+    },
```
cdata: Conventionally, Polymer Elements code style discourages this trailing comma.

iron-overlay-manager.html
```diff
@@ -12,60 +12,74 @@
...
+
+    /**
+     * Ensures that the minimum z-index of new overlays is at least `minimumZ`.
+     * This does not effect the z-index of any existing overlays.
+     *
+     * @param {number} minimumZ
+     */
+    ensureMinimumZ: function(minimumZ) {
+      this._minimumZ = Math.max(this._initialZIndex, minimumZ);
```
cdata: `_initialZIndex` does not seem to be initialized anywhere.

iron-overlay-manager.html
```diff
@@ -12,60 +12,74 @@
...
-    }
+    _applyOverlayZ: function(overlay, aboveZ) {
+      this._setZ(overlay, aboveZ + 2);
+    },
 
-    function setZ(element, z) {
+    _setZ: function(element, z) {
       element.style.zIndex = z;
-    }
+    },
```
cdata: Could you re-organize the public and protected implementation so that it is grouped together?

iron-overlay-manager.html
```diff
@@ -12,60 +12,74 @@
...
 
     // track overlays for z-index and focus managemant
-    function addOverlay(overlay) {
-      var z0 = currentOverlayZ();
-      overlays.push(overlay);
-      var z1 = currentOverlayZ();
-      if (z1 <= z0) {
-        applyOverlayZ(overlay, z0);
+    addOverlay: function(overlay) {
+      var minimumZ = Math.max(this.currentOverlayZ(), this._minimumZ);
```
cdata: It seems like the implementation of `currentOverlayZ` will default to returning `this._minimumZ`, so this `Math.max` may not be necessary?

cdata:
There are some small issues, but overall the approach looks good. Some tests would be very appreciated :bow:

iron-overlay-manager.html
justinfagnani: Bad push, it should have been `_minimumZ`.

iron-overlay-manager.html
justinfagnani: :rage:

iron-overlay-manager.html
justinfagnani: :ok:

iron-overlay-manager.html
justinfagnani: `currentOverlayZ` defaults to `_minimumZ` of there are no overlays, but if there are overlays it returns the z-index of the last opened overlay (I actually wonder about that behavior, but it's what was there). In cases where `ensureMinimumZ` is called after an overlay has been added, with a value greater than the current Z, then `currentOverlayZ` will return a value less than `_minimumZ`.

justinfagnani:
Ok, added tests and addressed comments.

cdata:
This looks great! Thanks for the fix :+1:

PR #53, opened 2015-11-18
stop bringing in the paper-styles kitchen sink

notwaldorf:
Fixes the generic usage of `paper-styles.html` and `/deep/` classes and uses the specific imports instead. Also updates the test to the new `wct#^3.4.0` way

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -19,14 +19,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
 
-    <link rel="import" href="../../iron-flex-layout/classes/iron-flex-layout.html">
-    <!-- <link rel="import" href="../core-transition/core-transition-css.html"> -->
+    <link rel="import" href="../../iron-flex-layout/iron-flex-layout.html">
     <link rel="import" href="simple-overlay.html">
-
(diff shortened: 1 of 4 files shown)


test/iron-overlay-behavior.html
```diff
@@ -20,10 +20,7 @@
     <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
 
     <script src="../../web-component-tester/browser.js"></script>
-    <script src="../../test-fixture/test-fixture-mocha.js"></script>
     <script src="../../iron-test-helpers/test-helpers.js"></script>
-
-    <link rel="import" href="../../test-fixture/test-fixture.html">
```
cdata: :golf:

demo/index.html
```diff
@@ -78,7 +82,7 @@
 
       <button data-dialog="scrolling-2">scrolling 2</button>
 
-      <simple-overlay id="scrolling-2" class="with-margin full-height layout vertical">
+      <simple-overlay id="scrolling-2" class="with-margin full-height vertical">
```
cdata: Clear float! Full height! Can't lose!

cdata:
Looks like we are having the `Polymer.Base.fire` problem in the tests :/ filed https://github.com/PolymerElements/iron-overlay-behavior/issues/54

LGTM :+1:

PR #61, opened 2015-12-07
Compute position on attached

### LINKED ISSUE #57: Tests fail due to `Polymer.Base.fire` misuse.
Polymer changes internal implementation of `Polymer.Base.fire`, so this method can no longer be called with a custom context.

### LINKED ISSUE #60: position calculation triggered too early
`iron-overlay-behavior` triggers position calculation too early (or `ready`), resulting in a wrong/offsetted rendering of the element. Instead, this should be done on `attached`.

valdrinkoshi:
Fixes #60 by waiting for element to be attached to call `this._openChanged`. Only on attached we can safely compute the element's position.
Fixes #57 by using `MockInteractions.tap` :balloon:

### DIFF
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -20,7 +20,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
 
     <script src="../../web-component-tester/browser.js"></script>
-    <script src="../../iron-test-helpers/test-helpers.js"></script>
+    <link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
     <link rel="import" href="test-overlay.html">
 
   </head>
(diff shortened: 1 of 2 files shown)


cdata:
Can we add a regression test for #60?

valdrinkoshi:
@cdata done, test added :tada:

test/iron-overlay-behavior.html
```diff
@@ -100,11 +100,22 @@
           assert.equal(getComputedStyle(overlay).display, 'none', 'overlay starts hidden');
         });
 
-        test('overlay open by default', function() {
+        test('overlay open by default', function(done) {
           overlay = fixture('opened');
-          runAfterOpen(overlay, function() {
+          overlay.addEventListener('iron-overlay-opened', function() {
```
valdrinkoshi: changed to this, because `runAfterOpen` is actually calling `overlay.open()` which defeats the purpose of this test

cdata:
LGTM :+1: Thanks for fixing the tests!

PR #63, opened 2015-12-08
esc closes only one overlay

### LINKED ISSUE #62: ESC Closing all dialogs when nested
_From @erichnascimento on October 8, 2015 0:43_

When key ESC is pressed in nested dialogs, all dialogs are closed instead of the highest dialog only.

_Copied from original issue: PolymerElements/paper-dialog#45_

valdrinkoshi:
Fixes #62 by stopping immediate propagation of the event so that only one overlay gets closed on ESC.


cdata:
LGTM :+1:

PR #64, opened 2015-12-08
IronOverlayBehavior no longer needs click events to bubble to determine when to close.

bicknellr:
Fixes PolymerElements/iron-menu-behavior#40.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -140,7 +140,6 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     listeners: {
-      'tap': '_onClick',
       'iron-resize': '_onIronResize'
     },
 
@@ -397,20 +396,10 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     _onCaptureClick: function(event) {
-      // attempt to close asynchronously and prevent the close of a tap event is immediately heard
-      // on target. This is because in shadow dom due to event retargetting event.target is not
-      // useful.
-      if (!this.noCancelOnOutsideClick && (this._manager.currentOverlay() == this)) {
-        this._cancelJob = this.async(function() {
-          this.cancel();
-        }, 10);
-      }
-    },
-
-    _onClick: function(event) {
-      if (this._cancelJob) {
-        this.cancelAsync(this._cancelJob);
-        this._cancelJob = null;
+      if (!this.noCancelOnOutsideClick &&
+          this._manager.currentOverlay() === this &&
+          Polymer.dom(event).path.indexOf(this) === -1) {
+        this.cancel();
       }
     },
 


cdata:
@bicknellr Looks like we need a rebase.

iron-overlay-behavior.html
```diff
@@ -393,20 +392,8 @@
...
-          this.cancel();
-        }, 10);
-      }
-    },
-
-    _onClick: function(event) {
-      if (this._cancelJob) {
-        this.cancelAsync(this._cancelJob);
-        this._cancelJob = null;
+      if (event.path.indexOf(this) === -1) {
```
cdata: What happened to the `this.noCancelOnOutsideClick` part? Also, I think `this._manager.currentOverlay() == this` is still relevant..

test/iron-overlay-behavior.html
```diff
@@ -181,7 +181,7 @@
             overlay.addEventListener('iron-overlay-canceled', function(event) {
               done();
             });
-            Polymer.Base.fire.call(document, 'click');
+            Polymer.Base.fire('click', null, {node: document});
```
cdata: We should use `MockInteractions` for this instead.

bicknellr:
Looks like `event.path` is supplied by Chrome, not Polymer...

cdata:
You should use `Polymer.dom(event).path` (I think this works.. I know we have something that populates the path).

test/iron-overlay-behavior.html
bicknellr: Fixed.

iron-overlay-behavior.html
bicknellr: Not sure why I did this; it's back in.

bicknellr:
Yup, `Polymer.dom(event).path` works!

iron-overlay-behavior.html
```diff
@@ -397,20 +396,10 @@
...
-          this.cancel();
-        }, 10);
-      }
-    },
-
-    _onClick: function(event) {
-      if (this._cancelJob) {
-        this.cancelAsync(this._cancelJob);
-        this._cancelJob = null;
+      if (Polymer.dom(event).path.indexOf(this) === -1 &&
```
cdata: This is a minor detail, but it would probably be ideal to check the event path last. I think computing the event path is relatively most expensive of the three booleans.

bicknellr: Makes sense; updated.

cdata:
LGTM :+1:

bastienarm:
Up,

it seems that there is another problem with this.

I've made you a JSBin

https://jsbin.com/lapedorogo/edit?html,console,output

I don't know why, but if you use paper-dropdown-menu + paper-menu + paper-item inside a dialog (dialog and paper-dropdown-menu are differents module), when selecting an item in the paper-dropdown-menu, an event will be fired which seems to be "iron-overlay-closed"

bastienarm:
Update : no need to select an item, clicking outside to hide the list will have the same effect

PR #68, opened 2015-12-10
set focus back to last focused element on close

### LINKED ISSUE #67: keep track of the last external focused element
When the last overlay is closed, focus should move back to the origin element.
E.g.
- button opens overlay
- click outside overlay closes overlay

Expectation: button has focus

### LINKED ISSUE #94: dialogs should restore focus to the previously focused element after closing
_From @anthonytordillos on October 20, 2015 6:9_

"the object with focus before the dialog box is opened should be saved. This will allow restoring focus to this element when the dialog box is closed."

Is this something paper-dialog-behavior can provide, or are consumers of the element required to handle it themselves?

http://www.w3.org/TR/wai-aria-practices/#modal_dialog

_Copied from original issue: PolymerElements/paper-dialog-behavior#32_

valdrinkoshi:
Fixes #94 by storing the external focused element in the manager. Added `restore-focus-on-close` property.

~~Depends on https://github.com/Polymer/polymer/pull/2717~~

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -128,6 +128,14 @@ context. You should place this element as a child of `<body>` whenever possible.
         value: document
       },
 
+      /**
+       * Set to true to enable restoring of focus when overlay is closed.
+       */
+      restoreFocusOnClose: {
+        type: Boolean,
+        value: false
+      },
+
       _manager: {
         type: Object,
         value: Polymer.IronOverlayManager
@@ -235,7 +243,7 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     ready: function() {
-      // with-backdrop need tabindex to be set in order to trap the focus.
+      // with-backdrop needs tabindex to be set in order to trap the focus.
       // If it is not set, IronOverlayBehavior will set it, and remove it if with-backdrop = false.
       this.__shouldRemoveTabIndex = false;
       // Used for wrapping the focus on TAB / Shift+TAB.
@@ -245,7 +253,7 @@ context. You should place this element as a child of `<body>` whenever possible.
 
     attached: function() {
       // Call _openedChanged here so that position can be computed correctly.
-      if (this._callOpenedWhenReady) {
+      if (this.opened) {
         this._openedChanged();
       }
       this._observer = Polymer.dom(this).observeNodes(this._onNodesChange);
@@ -315,7 +323,6 @@ context. You should place this element as a child of `<body>` whenever possible.
 
       // wait to call after ready only if we're initially open
       if (!this._overlaySetup) {
-        this._callOpenedWhenReady = this.opened;
         return;
       }
 
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -18,6 +18,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    */
   Polymer.IronOverlayManagerClass = function() {
     this._overlays = [];
+    // Used to keep track of the last focused node before an overlay gets opened.
+    this._lastFocusedNodes = [];
+
     /**
      * iframes have a default z-index of 100, so this default should be at least
      * that.
@@ -36,7 +39,42 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return this._backdropElement;
       }.bind(this)
     });
-  }
+
+    /**
+     * The deepest active element.
+     * returns {?Node} element the active element
+     */
+    this.deepActiveElement = null;
+    Object.defineProperty(this, 'deepActiveElement', {
+      get: function() {
+        var active = document.activeElement;
+        // document.activeElement can be null
+        // https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement
+        while (active && active.root && Polymer.dom(active.root).activeElement) {
+          active = Polymer.dom(active.root).activeElement;
+        }
+        return active;
+      }.bind(this)
+    });
+  };
+
+  /**
+   * If a node is contained in an overlay.
+   * @private
+   * @param {Node} node
+   * @returns {Boolean}
+   */
+  Polymer.IronOverlayManagerClass.prototype._isChildOfOverlay = function(node) {
+    while (node && node !== document.body) {
+      // Use logical parentNode, or native ShadowRoot host.
+      node = Polymer.dom(node).parentNode || node.host;
+      // Check if it is an overlay.
+      if (node && node.behaviors && node.behaviors.indexOf(Polymer.IronOverlayBehaviorImpl) !== -1) {
+        return true;
+      }
+    }
+    return false;
+  };
 
   Polymer.IronOverlayManagerClass.prototype._applyOverlayZ = function(overlay, aboveZ) {
     this._setZ(overlay, aboveZ + 2);
@@ -56,6 +94,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     if (newZ <= minimumZ) {
       this._applyOverlayZ(overlay, minimumZ);
     }
+    var element = this.deepActiveElement;
+    // If already in other overlay, don't reset focus there.
+    if (this._isChildOfOverlay(element)) {
+      element = null;
+    }
+    this._lastFocusedNodes.push(element);
   };
 
   Polymer.IronOverlayManagerClass.prototype.removeOverlay = function(overlay) {
@@ -63,6 +107,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     if (i >= 0) {
       this._overlays.splice(i, 1);
       this._setZ(overlay, '');
+
+      var node = this._lastFocusedNodes[i];
+      // Focus only if still contained in document.body
+      if (overlay.restoreFocusOnClose && node && Polymer.dom(document.body).deepContains(node)) {
+        node.focus();
+      }
+      this._lastFocusedNodes.splice(i, 1);
     }
   };
 
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -24,6 +24,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
     <link rel="import" href="test-overlay.html">
     <link rel="import" href="test-overlay2.html">
+    <link rel="import" href="test-buttons.html">
 
     <style is="custom-style">
       iron-overlay-backdrop {
@@ -111,6 +112,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         </test-overlay2>
       </template>
     </test-fixture>
+
+    <test-buttons id="buttons"></test-buttons>
+    <input id="focusInput" placeholder="focus input">
 
     <script>
       function runAfterOpen(overlay, callback) {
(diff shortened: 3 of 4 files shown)


valdrinkoshi:
@cdata, FYI please review.
About the activeElement, I've done some tests and noticed that even with nested elements `document.activeElement` will retrieve the correct element (click the button, and check the logs) http://jsbin.com/tocayo/2/edit?html,console,output

cdata:
Can you add a quick test case for this behavior to the test suite?

cdata:
Keep in mind that `document.activeElement` will be inconsistent between ShadowDOM and ShadyDOM.

valdrinkoshi:
True story! I updated it to use ownerRoot. This PR now depends on https://github.com/Polymer/polymer/pull/2717 landing

valdrinkoshi:
@cdata, updated tests! PTAL

iron-overlay-behavior.html
```diff
@@ -241,15 +242,22 @@
       this._toggleListeners();
 
       if (this.opened) {
+        var ownerRoot = Polymer.dom(this).getOwnerRoot();
+        var doc = ownerRoot ? Polymer.dom(ownerRoot) : document;
```
cdata: :beers:

cdata:
I'm kind of scared that the tests pass even though https://github.com/Polymer/polymer/pull/2717 is not merged. Do we need to add a test that covers the condition of an overlay inside a ShadowRoot?

valdrinkoshi:
@cdata tests updated, now it will fail until https://github.com/Polymer/polymer/pull/2717 gets merged

cdata:
Very nice :+1: looking forward to 2717 merging soon :)

valdrinkoshi:
@cdata I realized that `_lastFocusedElement` is already a property of `paper-dialog-behavior` (which extends `iron-overlay-behavior`) https://github.com/PolymerElements/paper-dialog-behavior/blob/master/paper-dialog-behavior.html#L89
I'll use a different name in here, maybe `_lastFocusedNode`

iron-overlay-behavior.html
```diff
@@ -241,15 +242,22 @@
...
 
       // async here to allow overlay layer to become visible.
       this._openChangedAsync = this.async(function() {
         // overlay becomes visible here
         this.style.display = '';
-        // force layout to ensure transitions will go
-        /** @suppress {suspiciousCode} */ this.offsetWidth;
+        // Force layout to ensure transitions will go. Set it to itself so it
+        // doesn't get removed by compilers.
+        this.offsetWidth = this.offsetWidth;
```
cdata: :+1:

test/iron-overlay-behavior.html
```diff
@@ -228,9 +248,7 @@
             overlay.addEventListener('iron-overlay-canceled', function(event) {
               done();
             });
-            fireEvent('keydown', {
-              keyCode: 27
-            }, document);
+            MockInteractions.pressAndReleaseKeyOn(document, 27);
```
cdata: :+1:

cdata:
LGTM when polymer/polymer#2727 lands :+1:

valdrinkoshi:
@cdata, I believe that if every overlay stores the `_lastFocusedNode` and tries to restore the focus to it when opened changes, this might create some issues for a11y (e.g. the reader will start reading something, stop, start reading the next focused element..)
I think restoring focus is a duty of `iron-overlay-manager`; it already has a method `focusOverlay`, so what if we rename it `restoreFocus` and create another method `storeFocus`? `focusOverlay` will be kept so we are compatible with previous versions (but will be deprecated).
The idea is that the manager should restore the focus only if there are no more overlays opened.

iron-overlay-behavior.html
```diff
@@ -155,13 +155,20 @@
...
+    },
+
     ready: function() {
+      this._lastFocusedNode = null;
       this._ensureSetup();
     },
 
     attached: function() {
       // Call _openedChanged here so that position can be computed correctly.
-      if (this._callOpenedWhenReady) {
```
valdrinkoshi: 1 variable less!

iron-overlay-manager.html
```diff
@@ -78,21 +78,6 @@
     this._minimumZ = Math.max(this._minimumZ, minimumZ);
   };
 
-  Polymer.IronOverlayManagerClass.prototype.focusOverlay = function() {
```
valdrinkoshi: this is not needed anymore since now each overlay keeps track of the lastFocusedNode before applying the focus.

valdrinkoshi:
Updated. The main change is to handle focus-related things into `_applyFocus`. Now each overlay keeps track of the `_lastFocusedNode` before moving the focus, and gives it back if closed. This allowed to remove `focusOverlay` method from `iron-overlay-manager`.
Still failing until `polymer#1.2.4` is available.
@cdata PTAL :)

valdrinkoshi:
@cdata, another update for this PR :) 
We don't need anymore to check the activeElement, but instead we track the focus changes in the manager, keeping a reference to the last focused element not contained in an overlay.
PTAL

valdrinkoshi:
One known limitation for the focus return is the following:
with ShadowDOM, the `focus` event on `document.body` won't be fired when we set the focus on an element defined in the `:host` (e.g. see here http://jsbin.com/xezapo/13/edit?html,output), so the `iron-overlay-manager` won't return the focus to those elements.
@cdata LMK if you think there is an easy solution for this, otherwise I'd defer it to a separate PR.

iron-overlay-manager.html
```diff
@@ -26,7 +26,38 @@
...
+    }
+  };
+
+  /**
+   * @private
+   * @description Checks if the array contains an overlay, and returns it.
+   * @param {path: [HTMLNode]} array obtained from event.path
+   * @returns {overlat: HTMLNode|undefined} the overlay
+   */
+  Polymer.IronOverlayManagerClass.prototype._overlayParent = function(path) {
```
cdata: `private` should be indicated with two leading `_` characters, e.g., `__overlayParent`

valdrinkoshi: ups, will fix

cdata:
We need to use a strategy similar to https://github.com/cdata/focus-observer, or else it is not really possible to track the last focused element correctly.

Also relevant is this Firefox bug regarding `relatedTarget` and tracking the previous focused element from the details of a new focus event: https://bugzilla.mozilla.org/show_bug.cgi?id=962251

I'm kind of on the fence about tracking the last focused element this way. As implemented, it will be wrong a lot of the time in ShadowDOM.

iron-overlay-manager.html
```diff
@@ -26,7 +26,38 @@
...
+    }
+  };
+
+  /**
+   * @private
+   * @description Checks if the array contains an overlay, and returns it.
+   * @param {path: [HTMLNode]} array obtained from event.path
+   * @returns {overlay: HTMLNode|undefined} the overlay
+   */
+  Polymer.IronOverlayManagerClass.prototype.__overlayParent = function(path) {
```
cdata: :+1:

test/iron-overlay-behavior.html
```diff
@@ -631,7 +645,64 @@
...
+                done();
+              });
+            });
+          });
+        });
+
+      });
+
+      function forceFocus(node) {
+        MockInteractions.focus(node);
```
valdrinkoshi: @cdata FYI, `MockInteraction.focus` doesn't actually focus the node :x

valdrinkoshi:
@cdata tests will fail on safari, and at this point I don't know why they do :disappointed: 
On Safari, even setting `no-autofocus` won't work (see this test https://github.com/PolymerElements/iron-overlay-behavior/pull/98/files)
....why? :sos:

valdrinkoshi:
@cdata tests are now fixed! I discovered that `autofocus` will steal the focus the first time the element is displayed, so we now read the `deepestActiveElement` right before toggling `display: visible` of the overlay.
Also, on IE 9+ there are some issues when the overlay is contained in an iframe, `document.activeElement` would return a non-HTMLElement, so now we also check for that.

Tests are failing on IE10, can have a further look at those..

valdrinkoshi:
@tjsavage FYI (IE10 related tests failing)

iron-overlay-behavior.html
```diff
@@ -167,8 +167,26 @@
...
 
+    /**
+     * The deepest active element.
+     * @type Node
+     */
+    get _deepestActiveElement() {
+      var active = document.activeElement;
+      // In IE9,10,11 document.activeElement would return a non-HTMLElement when
+      // the overlay is inside an iframe, hence we should check for that.
+      if (active instanceof HTMLElement) {
```
valdrinkoshi: Without this check, tests fail with this error:
```
Windows 8.1 IE 11        ✖ test/iron-overlay-behavior.html » basic overlay tests » close an overlay by clicking outside
  Object doesn't support property or method 'focus' (data:text/javascript;charset=utf-8,%0A%0A%2F**%0AUse%20%60Polymer.IronOverlayBehavior%60%20to%20implement%20an%20element%20that%20can%20be%20hidden%20or%20shown%2C%20and%20displays%0Aon%20top%20of%20other%20content.%20It%20includes%20an%20optional%20backdrop%2C%20and%20can%20be%20used%20to%20implement%20a%20variety%0Aof%20UI%20controls%20including%20dialogs%20and%20drop%20downs.%20Multiple%20
// [...]
```
So basically `overlay._deepestActiveElement` is called after `attached` for that particular test, and `document.activeElement` is not an HTMLElement on IE11...

@notwaldorf @cdata any thoughts? I don't really like to put this check here just because of the tests :x

test/iron-overlay-behavior.html
```diff
@@ -631,7 +637,88 @@
...
+          document.body.appendChild(focusable);
+          focusable.focus();
+          assert.equal(overlay._deepestActiveElement, focusable, 'input is handled');
+          document.body.removeChild(focusable);
+        });
+
+        test('_deepestActiveElement handles shadow dom', function (done) {
+          var focusable = fixture('focusables').$.focusable0;
+          // Wait for rendering before setting focus
+          setTimeout(function () {
```
valdrinkoshi: this timeout and the ones at line 676 and 692 are needed for tests on IE10.
I've tried to use `Polymer.dom.flush()`, `TestHelpers.flushAsynchronousOperations()` and `TestHelpers.forceXIfStamp(fixture)`, but no luck with any. 
This is the only way I could find to have the tests passing until we :skull: IE10 from our tests.

iron-overlay-behavior.html
valdrinkoshi: In IE10 tests
```javascript
var active = document.activeElement;
if (active instanceof HTMLElement) {
  //..
} else {
  console.log(active); // logs an Object { __domApi: { } }
}
```
:confounded: :question:

valdrinkoshi:
Yay! Fixes all tests :grin: 
I've rebased this commit on top of PR #104. The diff is in the last commit, or here https://github.com/PolymerElements/iron-overlay-behavior/compare/clean-tests...issue-67

iron-overlay-manager.html
```diff
@@ -36,7 +39,41 @@
         return this._backdropElement;
       }.bind(this)
     });
-  }
+
+    /**
+     * The deepest active element not contained in an overlay.
+     * returns {element: HTMLNode} the active element
+     */
+    this.deepestActiveElement = null;
```
valdrinkoshi: Should rename it to `deepActiveElement`

iron-overlay-manager.html
```diff
@@ -36,7 +39,41 @@
...
+    Object.defineProperty(this, 'deepestActiveElement', {
+      get: function() {
+        var active = document.activeElement;
+        // document.activeElement can be null
+        // https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement
+        while (active && active.root && Polymer.dom(active.root).activeElement) {
+          active = Polymer.dom(active.root).activeElement;
+        }
+        if (active && !this.__overlayContaining(active)) {
```
valdrinkoshi: the check `!this.__overlayContaining(active)` should not be done here

valdrinkoshi:
Tests are failing on IE10. Tried locally and they all pass:
ie10-tests

@azakus any ideas?

valdrinkoshi:
Tests are green again :tada: 
I've also renamed the manager's method to `_isChildOfOverlay` and optimized it to check for `behaviors` instead of using `deepContains`.
Assigning this to @notwaldorf

notwaldorf:
So I have a general comment about the `noRestoreFocus` property: it seems that this PR is introducing a breaking change (you're changing the behaviour of focus), and adding an option to undo that. The correct thing to do if you're thinking about introducing a breaking change is to have a property that turns this feature _on_, not off. So, it would have to be called something like `restoreFocus`, default of false.

This of course unless you think this is a bug fix and not a breaking change.

valdrinkoshi:
You're right @notwaldorf, safer to have `restoreFocus`. Will update

valdrinkoshi:
Updated to a more descriptive `restore-focus-on-close`

iron-overlay-behavior.html
```diff
@@ -128,6 +128,14 @@
         value: document
       },
 
+      /**
+       * Set to true to enable restoring of focus when overlay is closed.
+       */
+      restoreFocusOnClose: {
```
notwaldorf: Why not just `restore-focus`?

valdrinkoshi: I thought it would make it clearer when/how this should be used, like for `noCancelOnEscKey` or `noCancelOnOutsideClick`

iron-overlay-manager.html
```diff
@@ -36,7 +39,42 @@
...
       }.bind(this)
     });
-  }
+
+    /**
+     * The deepest active element.
+     * returns {?Node} element the active element
+     */
+    this.deepActiveElement = null;
+    Object.defineProperty(this, 'deepActiveElement', {
```
notwaldorf: Does this need to be a getter, or can it just be a function?

notwaldorf:
LGTM, with one small question :)

iron-overlay-manager.html
valdrinkoshi: I made it a getter to "seem" more similar to `document.activeElement`, other than that no particular reason.

PR #71, opened 2015-12-15
cleanup _squelchNextResize

valdrinkoshi:
Fixes https://github.com/PolymerElements/paper-dialog/issues/68 and https://github.com/PolymerElements/paper-dialog/issues/77
`iron-overlay-behavior` needs to refit itself only once when opened.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -226,6 +226,7 @@ context. You should place this element as a child of `<body>` whenever possible.
         this.removeAttribute('aria-hidden');
       } else {
         this.setAttribute('aria-hidden', 'true');
+        Polymer.dom(this).unobserveNodes(this._observer);
       }
 
       // wait to call after ready only if we're initially open
@@ -247,8 +248,9 @@ context. You should place this element as a child of `<body>` whenever possible.
       this._openChangedAsync = this.async(function() {
         // overlay becomes visible here
         this.style.display = '';
-        // force layout to ensure transitions will go
-        /** @suppress {suspiciousCode} */ this.offsetWidth;
+        // Force layout to ensure transition will go. Set offsetWidth to itself
+        // so that compilers won't remove it.
+        this.offsetWidth = this.offsetWidth;
         if (this.opened) {
           this._renderOpened();
         } else {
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
@notwaldorf @cdata kindly review

iron-overlay-behavior.html
```diff
@@ -341,9 +341,7 @@
       }
 
       this.fire('iron-overlay-opened');
-
-      this._squelchNextResize = true;
-      this.async(this.notifyResize);
+      this.async(this.refit);
```
cdata: Does refit cause `notifyResize` to be called?

valdrinkoshi: no, if that was the case the test i added would fail

cdata: `notifyResize` has additional side-effects. Children of the overlay have their resize handlers called as well. So, even if `refit` should only be called once, `notifyResize` still needs to be called to maintain implementation parity.

valdrinkoshi: True. But when an overlay gets opened, it is not resized, it just needs to refit.
If we need to `notifyResize`, it might be that other listeners are invoked before the overlay.

What should happen is:
- overlay refits
- overlay fires notify resize
- other listeners are execute

If this order is not respected, other listeners might be executed too early, before the overlay gets a chance to refit.

WDYT? do we need to `notifyResize`?

valdrinkoshi:
@cdata, I'm now using `observeNodes` to properly call `notifyResize`. PTAL

valdrinkoshi:
Tests are failing because on microsoft edge it fails to maximize the window, hence `iron-resize` will be triggered once at the beginning because of that...lol!
@azakus FYI

iron-overlay-behavior.html
```diff
@@ -322,28 +324,14 @@
       this._finishRenderClosed();
     },
 
-    _onTransitionend: function(event) {
```
valdrinkoshi: dead code cleanup

cdata: :100:

cdata:
Hrm. I wonder about the performance consequences of observing nodes to trigger resize.

I guess we'll find out!

LGTM :+1:

PR #72, opened 2015-12-15
update backdrop parent to make it sibling of the opened overlay with backdrop

valdrinkoshi:
Fixes https://github.com/PolymerElements/paper-dialog/issues/44
`iron-overlay-backdrop` now is sibling of the top overlay with backdrop. 
Its `position: fixed` ensures it covers the whole screen and is not cropped. 
Custom styles will be applied if the css selector does not assume `iron-overlay-behavior` is a direct child of `body` :

```css
<style is="custom-style">

  /* will work fine */
  body iron-overlay-backdrop {
    --iron-overlay-backdrop: {
      backdround: red;
  }

  /* will NOT work */
  body > iron-overlay-behavior {
    --iron-overlay-backdrop: {
      background: blue;
  }
</style>

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -120,7 +120,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       <button data-dialog="backdrop">backdrop</button>
 
       <simple-overlay id="backdrop" with-backdrop>
-        Hello world!
+        <p>Hello world!</p>
+        <button data-dialog="backdrop2">backdrop2</button>
+        <simple-overlay id="backdrop2" with-backdrop>
+          backdrop 2
+        </simple-overlay>
       </simple-overlay>
 
       <button data-dialog="autofocus">autofocus</button>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by valdrinkoshi** [comment]
@cdata kindly review

**Comment by cdata** [comment]
@valdrinkoshi I'm a little worried about this implementation change. If the backdrop shows up in different trees, this means that things like positional CSS selectors will start to break.

**Comment by valdrinkoshi** [comment]
I made a slight optimization by making sure `iron-overlay-backdrop` gets moved before the `sample-dialog` (using `insertBefore`).

In any case: yes, this might break some selectors that heavily rely on pseudo selectors like `:nth-child`, e.g. this selector would fail:
```html
<style>
  div > :first-child(2) {
    border: 1px solid black;
  }
</style>

<div>
  <sample-dialog with-backdrop opened>
    dialog
  </sample-dialog>
  <p>
    hello <span>world</span>
  </p>
</div>
```

valdrinkoshi:
@cdata, will close this as i discovered that it is caused by a css property that `paper-drawer-panel` sets when in narrow mode

Issue #73, opened 2015-12-16
The backdrops of two different elements implementing iron-overlay-behavior are interfering


zinkkrysty:
If two different elements are created, `element-overlay-with-backdrop` and `element-overlay-with-backdrop2`, both implement `iron-overlay-behavior` and both used `with-backdrop`, opening one after another and closing them, one overlay closes the other one's backdrop, and one backdrop doesn't close (is orphaned).

I have a suspicion that the manager is not working as it should - the current tests only cover the case where the overlays share the same backdrop - but once we introduce a different element which does not share (the backdrop is created on `register` event) - there are issues with correctly attributing backdrops and closing the right one.

PR #74, opened 2015-12-16
Fix closing of backdrops of different overlay elements & add tests

zinkkrysty:
Bug described in #73

Moved responsability of tracking the backdrop and the active overlays with backdrops to the backdrop element.

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -81,7 +81,15 @@ Custom property | Description | Default
       _manager: {
         type: Object,
         value: Polymer.IronOverlayManager
-      }
+      },
+
+      /**
+       * Manages the active overlays using this backdrop
+       */
+      _activeOverlays: {
+        type: Array,
+        value: []
+      },
 
     },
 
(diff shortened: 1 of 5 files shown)


zinkkrysty:
I think I'm over my head here trying to fix this. I got stuck on testing if the backdrop is actually added to the DOM. I'm trying to add this:

```js
assert.isTrue(document.getElementsByTagName('iron-overlay-backdrop').length === 1, 'there is a backdrop in the DOM tree');
```

in the `runAfterOpen` callback, but it's failing to recognize the backdrop in the DOM, and I can't see it either in the element inspector.

Anyone willing to assist?

valdrinkoshi:
if you try to do `elemet.tagName` you'll see that in some browsers it will be capitalized, so is better to use `document.querySelectorAll` instead of `document.getElementsByTagName`

valdrinkoshi:
About the fix, I'm not sure we should be adding multiple `iron-overlay-backdrop` elements. 
Instead, we should keep only one for all the `iron-overlay-behavior` elements. So maybe we should generate it in the `iron-overlay-manager` instead of `iron-overlay-behavior`.
@cdata WDYT?

zinkkrysty:
You are right, I realized my approach is not that good. I opened #79 with just the failing tests.

PR #75, opened 2015-12-17
Convert IronOverlayManager to a singleton class

rictic:


cdata:
This is not a behavior. Is this annotation misleading?

rictic:
Oh, it's not? Is it just a singleton object?

zinkkrysty:
I think that the manager violates the SRP, as it both managing the overlay and the backdrops. The backdrop managing is not necessary since backdrops are shared only in some cases (when the element used is the same). But when 2 elements implement the iron-overlay-behavior, the backdrop isn't shared, instead they use 2 different backdrop elements. See #73 for more details.

That's why I believe that each backdrop should manage its own overlays and not let the manager do that.

rictic:
Converted it from an object literal to a class with a single instantiation, PTAL.

@zinkkrysty this PR is just to explain what IronOverlayManager is doing to closure compiler. Please open an issue to debate the design.

cdata:
@rictic @zinkkrysty :+1: Please open an issue to discuss!

cdata:
@rictic this looks fine to me, but please co-ordinate this change with @valdrinkoshi since he has one or two PRs open that touch the same code.

valdrinkoshi:
LGTM too
@rictic I'll merge this one and solve the conflicts on my side

rictic:
Thanks!

PR #79, opened 2015-12-22
Bugfix: backdrop not shared in different elements, causing inconsistent closing

zinkkrysty:
Bug described in #73. This pull request adds the failing tests. 

My approach in #74 is flawed and therefore re-created the tests without the solution. Let's discuss what a reasonable approach to the solution would be. I think that @valdrinkoshi is right, we shouldn't have multiple backdrops, we should only allow one backdrop and that should be managed by the `iron-overlay-manager`.

**EDIT** I've now commited the solution, much more elegant and straightforward: don't allow more than 1 instance of `iron-overlay-backdrop` and let the manager handle the creation of that.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -148,15 +148,11 @@ context. You should place this element as a child of `<body>` whenever possible.
      * @type Node
      */
     get backdropElement() {
-      return this._backdrop;
+      return this._manager.backdropElement;
     },
 
     get _focusNode() {
       return Polymer.dom(this).querySelector('[autofocus]') || this;
-    },
-
-    registered: function() {
-      this._backdrop = document.createElement('iron-overlay-backdrop');
     },
 
     ready: function() {
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -107,6 +107,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   };
 
+  Object.defineProperty(Polymer.IronOverlayManagerClass.prototype, "backdropElement", {
+    get: function() {
+      if (!this._backdropElement) {
+        this._backdropElement = document.createElement('iron-overlay-backdrop');
+      }
+      return this._backdropElement;
+    }
+  });
+
   Polymer.IronOverlayManagerClass.prototype.getBackdrops = function() {
     return this._backdrops;
   };
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -22,6 +22,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <script src="../../web-component-tester/browser.js"></script>
     <link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
     <link rel="import" href="test-overlay.html">
+    <link rel="import" href="test-overlay2.html">
 
   </head>
   <body>
@@ -76,6 +77,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         <test-overlay with-backdrop class="overlay-3">
           Overlay 3 with backdrop
         </test-overlay>
+      </template>
+    </test-fixture>
+
+    <test-fixture id="backdrop-different-elements">
+      <template>
+        <test-overlay with-backdrop class="overlay-1">
+          Overlay 1 with backdrop
+        </test-overlay>
+        <test-overlay2 with-backdrop class="overlay-2">
+          Overlay 2 with backdrop
+        </test-overlay2>
       </template>
     </test-fixture>
 
@@ -435,6 +447,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             Polymer.dom(backdrop.parentNode).removeChild(backdrop);
             Polymer.dom.flush();
             assert.isNull(backdrop.parentNode, 'backdrop is removed from DOM');
+            assert.lengthOf(document.querySelectorAll('iron-overlay-backdrop'), (0));
             done();
           });
         });
@@ -498,6 +511,44 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         });
 
       });
+
+      suite('multiple overlays with backdrop implemented in different elements', function () {
+        var overlays;
+
+        setup(function() {
+          overlays = fixture('backdrop-different-elements');
+        });
+
+        test('multiple overlays share the same backdrop', function() {
+          assert.equal(overlays[0].backdropElement, overlays[1].backdropElement);
+        });
+
+        test('when overlays close, the backdrop is closed', function(done) {
+          runAfterOpen(overlays[0], function () {
+            assert.lengthOf(document.querySelectorAll('iron-overlay-backdrop'), 1);
+
+            // After second overlay is closed, both backdrops should be hidden
+            overlays[1].addEventListener('iron-overlay-closed', function() {
+              Polymer.Base.async(function () {
+                assert.isFalse(overlays[1].backdropElement.opened, 'second overlay backdrop is closed');
+                assert.isFalse(overlays[0].backdropElement.opened, 'first overlay backdrop is closed');
+                done();
+              }, 1);
+            });
+            // After second overlay is opened, immediately close it
+            overlays[1].addEventListener('iron-overlay-opened', function() {
+              Polymer.Base.async(function () {
+                overlays[1].close();
+              }, 1);
+            });
+
+            // Immediately close first overlay and open the other one
+            overlays[0].close();
+            overlays[1].open();
+          });
+        });
+
+      })
 
       suite('a11y', function() {
 
--- a/test/test-overlay2.html
+++ b/test/test-overlay2.html
@@ -1,0 +1,49 @@
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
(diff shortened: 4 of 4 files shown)


iron-overlay-behavior.html
```diff
@@ -155,11 +155,8 @@
       return Polymer.dom(this).querySelector('[autofocus]') || this;
     },
 
-    registered: function() {
-      this._backdrop = document.createElement('iron-overlay-backdrop');
-    },
-
     ready: function() {
+      this._backdrop = this._manager.getBackdropInstance();
```
valdrinkoshi: you could completely remove `this._backdrop` from here, and just return `this.manager.getBackdropInstance()` in `get backdropElement()`

iron-overlay-manager.html
```diff
@@ -106,6 +106,13 @@
       }
     },
 
+    getBackdropInstance: function() {
```
valdrinkoshi: you could make this a getter, in the same fashion as in `iron-overlay-behavior`
```javascript
// ...
_backdropInstance: document.createElement('iron-overlay-backdrop'),

get backdropElement() {
   return this._backdropInstance;
},
// ...

File: test/iron-overlay-behavior.html [hunk 3]
```diff
@@ -421,6 +434,45 @@
...
+
+        setup(function() {
+          overlays = fixture('backdrop-different-elements');
+        });
+
+        test('multiple overlays share the same backdrop', function() {
+          assert.equal(overlays[0].backdropElement, overlays[1].backdropElement);
+        });
+
+        test('multiple overlays close the corrent backdrop', function(done) {
```
valdrinkoshi: small typo, "corrent" should be "current"

zinkkrysty: I actually meant "correct". But since we are only talking about one single backdrop now maybe it should just be "the backdrop"

iron-overlay-manager.html
zinkkrysty: great idea!

test/iron-overlay-behavior.html
```diff
@@ -421,6 +434,45 @@
...
+          runAfterOpen(overlays[0], function () {
```
valdrinkoshi: Consider adding a method `runAfterClose` which is the same as `runAfterOpen`, but just adds a listener for `iron-overlay-close` instead of `iron-overlay-opened`.
Also, notice that `runAfterOpen` will execute the callback after 1ms.
With that this code could become something like this:
```javascript
runAfterOpen(overlays[0], function() {
  overlays[0].close();
  runAfterOpen(overlays[1], function() {
    runAfterClose(overlays[1], function() {
       // your tests here!
    });
  });
});

File: iron-overlay-manager.html (hunk 2 above)
Comment (zinkkrysty) [inline iron-overlay-manager.html]: I'm just wondering, if it's slightly more efficient to only create the backdrop element in case we need it? lazy-load vs eager-load ...

File: test/iron-overlay-behavior.html (hunk 3 above)
Comment (valdrinkoshi) [inline test/iron-overlay-behavior.html]: fair enough :+1:

File: iron-overlay-manager.html (hunk 2 above)
Comment (valdrinkoshi) [inline iron-overlay-manager.html]: you're right, you can do the lazy loading within the `get backdropElement()`

File: iron-overlay-behavior.html [hunk 5]
```diff
@@ -155,11 +155,8 @@
       return Polymer.dom(this).querySelector('[autofocus]') || this;
     },
 
-    registered: function() {
-      this._backdrop = document.createElement('iron-overlay-backdrop');
```
valdrinkoshi: we should also move the import to `iron-overlay-manager` (line 14)
```html
<link rel="import" href="iron-overlay-backdrop.html">

Comment (zinkkrysty) [inline iron-overlay-behavior.html]: makes sense I'll do that and amend the previous commit with fixes

Comment (valdrinkoshi) [inline iron-overlay-behavior.html]: oh, and we can also remove it from here (the line 14), since the import of `iron-overlay-manager` will import the backdrop :)

File: iron-overlay-manager.html [hunk 6]
```diff
@@ -22,6 +23,8 @@
 
     _backdrops: [],
 
+    _backdropElement: document.createElement('iron-overlay-backdrop'),
```
valdrinkoshi: @zinkkrysty tests are failing because of this line. If you do lazy loading on the getter solves it

valdrinkoshi: the reason why this is failing is because the manager is a singleton, an object, and here we add a key. To access to it we should do `Polymer.IronOverlayManager._backdropElement` in the getter (ugly).
Much better to lazy load by checking if `this._backdropInstance` exists.
```javascript
   get backdropElement() {
      if (!this._backdropElement) {
        this._backdropElement = document.createElement('iron-overlay-backdrop');
      }
      return this._backdropElement;
    },

Comment (zinkkrysty) [inline iron-overlay-manager.html]: thanks I was trying to run the tests locally with wct - that thing doesn't want to work with my setup, but you've beat me to it figuring out the issue. I'll make this fix and commit once more

File: iron-overlay-behavior.html (hunk 5 above)
Comment (zinkkrysty) [inline iron-overlay-behavior.html]: I thought I did that, but somehow I didn't commit :) I'll do it now.

File: iron-overlay-manager.html (hunk 6 above)
Comment (zinkkrysty) [inline iron-overlay-manager.html]: Hmm there is a problem now, a circular dependency issue I think: the manager imports the backdrop and the backdrop imports the manager, I guess that's why I get `Uncaught TypeError: Cannot read property 'getBackdrops' of undefined`

Comment (zinkkrysty) [inline iron-overlay-manager.html]: Yeah I can confirm it. If I move the `backdrop` import back into `iron-overlay-behavior` the tests pass

Comment (valdrinkoshi) [inline iron-overlay-manager.html]: Oh bummer...
Well, thinking of it again, it makes sense for `iron-overlay-behavior` to do the import, after all he's the one that uses the backdropElement's methods. Sorry for the confusion, my previous suggestion introduced this issue :disappointed:

File: iron-overlay-behavior.html (hunk 5 above)
Comment (valdrinkoshi) [inline iron-overlay-behavior.html]: this was the suggestion that causes the tests to fail :x
Never mind it, the import should be done here, in `iron-overlay-behavior` since in here we use the methods of `iron-overlay-backdrop`, and should not be done in `iron-overlay-manager`

File: iron-overlay-manager.html (hunk 6 above)
Comment (zinkkrysty) [inline iron-overlay-manager.html]: Well that's fine, it made sense for me as well. But I think this is a bit of code smell. I mean there shouldn't be circular dependency between 2 elements, should there?

**Comment by zinkkrysty** [comment]
Thanks for the suggestions and the help @valdrinkoshi , it's nice to see this bug finally squashed :) :tada:

File: iron-overlay-behavior.html [hunk 7]
```diff
@@ -11,8 +11,8 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../iron-fit-behavior/iron-fit-behavior.html">
 <link rel="import" href="../iron-resizable-behavior/iron-resizable-behavior.html">
-<link rel="import" href="iron-overlay-backdrop.html">
 <link rel="import" href="iron-overlay-manager.html">
+<link rel="import" href="iron-overlay-backdrop.html">
```
valdrinkoshi: would you mind reverting this as before? :angel: 
(since there is no particular reason to be moved as last import)

zinkkrysty: hmm, sure. My reasoning was that the backdrop needs the manager before but if I think about it, the backdrop actually imports the manager, so should be fine. I'll do that now.

valdrinkoshi:
Awesome job @zinkkrysty! I've let a minor comment, & I have a last request: could you squash your commits into one (so we have a clean history in our repo) ?

Overall LGTM, I defer the final :+1: and merge to @cdata :)

zinkkrysty:
Ok thanks! I'll squash the commits. I wasn't sure if you needed the history

valdrinkoshi:
Thanks @zinkkrysty, LGTM :+1: 
@cdata, ball in your court :soccer:

valdrinkoshi:
@zinkkrysty would you mind rebase this one? For that getter on `iron-overlay-manager`, you can use [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)

zinkkrysty:
Sure, I'll give it a go now

zinkkrysty:
Bummer, I've implemented the getter like this:

```
  Object.defineProperty(Polymer.IronOverlayManagerClass.prototype, "backdropElement", {
    get: function() {
      if (!this._backdropElement) {
        this._backdropElement = document.createElement('iron-overlay-backdrop');
      }
      return this._backdropElement;
    }
  });
```

But get a failing test at: 

```
assert.lengthOf(document.querySelectorAll('iron-overlay-backdrop'), 0); // It's 1 instead
```

Meaning that somehow the backdrop is not removed from the document when the overlays are closed.

I can push to this branch, but I already rebased and I wanted to show that the tests were indeed passing and my fix was working previously.

zinkkrysty:
Could the removal of the `completeBackdrop` method be the source of this regression? I guess that was the one that was removing the backdrop once it wasn't used anymore.

https://github.com/PolymerElements/iron-overlay-behavior/commit/61928b26e72d1771c6d96021cb04c062599d2504#diff-5e1626c54b6ff0c467b99e2307ad8af4L352

zinkkrysty:
Ah, of course, the new code closes the backdrop after transition ends, that's why if I add a timeout of 1s to the test, it passes. 

Any ideas on how to check the presence of the backdrop in the DOM after the `transitionEnd` callback, or do you think it's better to remove that assertion, as it's being tested for elsewhere?

valdrinkoshi:
You can manually call:
```javascript
overlay[1].close();
overlay[1].backdropElement.complete();
```
So that you don't need to wait for animations to be over

zinkkrysty:
Ok, I guess that defeats the purpose. I wanted to make sure that `complete()` is called after the backdrops are closed, but that is probably subject for another test. I'll remove that line and push the changes.

cdata:
This LGTM :+1: Thanks for the contribution!

PR #80, opened 2015-12-22
update backdrop on with-backdrop changes

### LINKED ISSUE #78: with-backdrop changes should update backdropElement
When an `iron-overlay-behavior` element is open, changes to `with-backdrop` should update the backdropElement presence.

E.g.
```html
<simple-overlay with-backdrop opened>
  Hello
</simple-overlay>
<script>
  setTimeout(function() {
    var overlay = document.querySelector('simple-overlay');
    overlay.withBackdrop = false; 
    // the backdrop element should have been closed
    console.log(overlay.backdropElement.opened); // should log false
  }, 1000);
</script>

### PR DESCRIPTION
**Pull request by valdrinkoshi**
Fixes #78 by adding an observer for `with-backdrop` changes.

`iron-overlay-manager` has a specific method to get the `z-index` for backdrop. This is done to make sure that backdrop's index is just below the first opened overlay that has `with-backdrop = true`. Also, its `trackBackdrop()` method now checks for duplicates and ensures that the passed element is opened and has a backdrop.

`iron-overlay-backdrop` will handle the call to `complete()` right after the transition ends. In addition, it updates its `z-index` on `prepare()` and on `close()`.

`iron-overlay-behavior` has been cleaned accordingly.

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -85,13 +85,18 @@ Custom property | Description | Default
 
     },
 
+    listeners: {
+      'transitionend' : '_onTransitionend'
+    },
+
     /**
      * Appends the backdrop to document body and sets its `z-index` to be below the latest overlay.
      */
     prepare: function() {
+      // Always update z-index
+      this.style.zIndex = this._manager.backdropZ();
       if (!this.parentNode) {
         Polymer.dom(document.body).appendChild(this);
-        this.style.zIndex = this._manager.currentOverlayZ() - 1;
       }
     },
 
@@ -109,9 +114,14 @@ Custom property | Description | Default
      * Hides the backdrop if needed.
      */
     close: function() {
+      // Always update z-index
+      this.style.zIndex = this._manager.backdropZ();
       // only need to make the backdrop invisible if this is called by the last overlay with a backdrop
       if (this._manager.getBackdrops().length < 2) {
         this._setOpened(false);
+        // complete() will be called after the transition is done.
+        // If animations are disabled via custom-styles, user is expected to call
+        // complete() after close()
       }
     },
 
@@ -122,6 +132,12 @@ Custom property | Description | Default
       // only remove the backdrop if there are no more overlays with backdrops
       if (this._manager.getBackdrops().length === 0 && this.parentNode) {
         Polymer.dom(this.parentNode).removeChild(this);
+      }
+    },
+
+    _onTransitionend: function (event) {
+      if (event && event.target === this) {
+        this.complete();
       }
     }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: iron-overlay-behavior.html [hunk 1]
```diff
@@ -296,14 +312,14 @@
...
       this._manager.addOverlay(this);
 
-      if (this.withBackdrop) {
-        this.backdropElement.prepare();
-        this._manager.trackBackdrop(this);
-      }
-
       this._preparePositioning();
       this.fit();
       this._finishPositioning();
```
valdrinkoshi: these lines about positioning will update the element's style, which in some cases influences the calculation of `z-index`. Executing these before the backdrop updates fixes the issue.

valdrinkoshi:
@cdata kindly review

iron-overlay-backdrop.html
```diff
@@ -109,9 +114,14 @@
...
      */
     close: function() {
+      // Always update z-index
+      this.style.zIndex = this._manager.backdropZ();
       // only need to make the backdrop invisible if this is called by the last overlay with a backdrop
       if (this._manager.getBackdrops().length < 2) {
         this._setOpened(false);
+        // complete() will be called after the transition is done.
+        // If animations are disabled via custom-styles, user is expected to call
+        // complete() after close()
```
valdrinkoshi: @cdata potentially we could have a property for the backdrop `noAnimation`, should I do it here or in a separate PR?

cdata: Let's do a separate PR. Could you file an issue to describe the enhancement?

iron-overlay-behavior.html
```diff
@@ -80,8 +80,8 @@
        * Set to true to display a backdrop behind the overlay.
        */
       withBackdrop: {
-        type: Boolean,
-        value: false
+        observer: '_withBackdropChanged',
+        type: Boolean
```
cdata: No more default value?

iron-overlay-manager.html
```diff
@@ -30,6 +30,31 @@
...
+    },
+
+    /**
+     * Calculates the minimum z-index for the overlay.
+     */
+    _getOverlayZ: function (overlay) {
+      var z = this._minimumZ;
+      if (overlay) {
+        var z1 = window.getComputedStyle(overlay).zIndex;
+        if (!isNaN(z1)) {
```
cdata: There are some [potentially undesired coercion effects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior) when using `isNaN`. They may not be an issue here (since the string is typically coming from a style, not from user input), but worth keeping in mind.

cdata:
Minor feedback, otherwise LGTM :+1:

iron-overlay-behavior.html
valdrinkoshi: we never really check if is `false`, so it should be a bit more performant to have one setter less.

valdrinkoshi:
@cdata, updated the check for `isNaN`, but since `Number.isNaN` is not supported in IE, I've used a direct check `z1 === z1` & let a comment there
Merging this

PR #82, opened 2016-01-06
code cleanup & better handling event listeners 

### LINKED ISSUE #27: closingReason.canceled not set correctly after closing overlay 2nd time

valdrinkoshi:
Fixes #27  by making sure `canceled` is set before updating `opened` (so that `closingReason` will be properly setup). 
Remove setting default value for booleans `canceled, withBackdrop, noAutoFocus, noCancelOnEscKey, noCancelOnOutsideClick`. 
Remove event listeners immediately after `opened` changes, and add them asynchronously if `opened = true`.
`_onCaptureKeydown` will not stop the event propagation but instead check if it is the current overlay (ensures the top one is closed)

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -180,6 +180,7 @@ context. You should place this element as a child of `<body>` whenever possible.
      * Toggle the opened state of the overlay.
      */
     toggle: function() {
+      this._setCanceled(false);
       this.opened = !this.opened;
     },
 
@@ -187,16 +188,16 @@ context. You should place this element as a child of `<body>` whenever possible.
      * Open the overlay.
      */
     open: function() {
+      this._setCanceled(false);
       this.opened = true;
-      this.closingReason = {canceled: false};
     },
 
     /**
      * Close the overlay.
      */
     close: function() {
-      this.opened = false;
       this._setCanceled(false);
+      this.opened = false;
     },
 
     /**
@@ -208,8 +209,8 @@ context. You should place this element as a child of `<body>` whenever possible.
         return;
       }
 
-      this.opened = false;
       this._setCanceled(true);
+      this.opened = false;
     },
 
     _ensureSetup: function() {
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
@cdata kindly review.
Tests are failing and it is related to resizing issues. This should be fixed once https://github.com/PolymerElements/iron-overlay-behavior/pull/71 gets merged.

iron-overlay-behavior.html
```diff
@@ -72,16 +72,14 @@
       canceled: {
         observer: '_canceledChanged',
         readOnly: true,
-        type: Boolean,
-        value: false
```
cdata: Could you go into some detail about why we are removing all of the default values?

valdrinkoshi:
@cdata, I've removed the default values of booleans mainly to save some unnecessary settings, with the idea that performance will be improved. Right now not all the properties have observers.
`undefined` and `false` have the same effect for boolean properties anyways. We should probably have an automatic setting of booleans to `false` done by polymer instead of doing it in each element, WDYT?

iron-overlay-behavior.html
cdata: Bump.

iron-overlay-behavior.html
```diff
@@ -115,7 +111,10 @@
       closingReason: {
         // was a getter before, but needs to be a property so other
         // behaviors can override this.
-        type: Object
+        type: Object,
+        value: {
```
cdata: This needs to be...

```
value: function() {
  return { canceled: false };
}
```

Or else all instances will share the same default value instance.

cdata:
@valdrinkoshi Removing the default values is probably fine in a lot of cases, but even if no-one is observing them in our code, it represents breaking API changes.

iron-overlay-behavior.html
valdrinkoshi: This previously wasn't set because `canceled` had a default value and its observer would update the `closingReason`.
If I let the default value of `canceled` to false, then this should be reverted.

valdrinkoshi:
@cdata understood, I guess is better to postpone the change of default values for now. I'll update the PR so it doesn't modify the default values

valdrinkoshi:
@cdata PR updated, ready for another :eyes:

cdata:
LGTM :+1:

PR #84, opened 2016-01-11
immediately track backdrop on opened changes

valdrinkoshi:
Fixes https://github.com/PolymerElements/paper-dialog/issues/77 by updating the manager's `backdrops` array on open changed.
Without this fix, when we add transitions to an overlay, it might be that they take longer than the animation of the backdrop. The manager needs to be updated as soon as possible so that once the backdrop is done, it can safely rely on the `manager.getBackdrops()` to be of the right length


valdrinkoshi:
@cdata kindly review

cdata:
LGTM :+1:

PR #86, opened 2016-01-12
add backdrop class to parent, use css pseudo-class for rendering

valdrinkoshi:
Fixes https://github.com/PolymerElements/paper-dialog/issues/79, fixes https://github.com/PolymerElements/paper-dialog/issues/44
Based on the idea started in https://github.com/PolymerElements/iron-overlay-behavior/pull/72
The handling of the backdrop is moved to the `iron-overlay-manager`; the `iron-overlay-backdrop` class will be added to the overlay's parent to handle the backdrop rendering. We use `::after` css selector, so positioning selectors will continue to work.
Manager will handle up to 99 overlays without backdrop (see the minimumZ code), after which the correct z-indexing is not guaranteed (for now).

Styling for the `iron-overlay-backdrop` still works as expected:
```html
<style is="custom-style">
  :root {
    --iron-overlay-backdrop-background-color: red;
    --iron-overlay-backdrop-opacity: 0.2;
    --iron-overlay-backdrop: {
      box-sizing: border-box;
      border: 20px solid blue;
    }
  }
</style>
```

`paper-dialog-behavior` needs to be updated too.


valdrinkoshi:
@cdata, the disadvantage of this approach is that if there is already a selector that operates on the parent's `::after`, this might override it: 
http://jsbin.com/kiviru/4/edit?html,output

Another approach is to use `outline` `::after` on the overlay itself:
http://jsbin.com/tojaza/2/edit?html,output

Both of these have issues with animations e.g. if there are other transitions on the elements we touch.

Another approach is keep the `backdropElement` and add it to the overlay itself (and use `outline` for the `backdropElement`):
http://jsbin.com/nowavew/5/edit?html,output

valdrinkoshi:
Ok, I think I found a better way to handle this; the basic idea is to use css to create a "hole" in the `backdropElement` in order for the user to interact in that box. This is the less disruptive way i could come up with:
https://github.com/PolymerElements/iron-overlay-behavior/compare/backdrop-refit

Closing this PR, and open soon the other

PR #89, opened 2016-01-16
fix bug 'indexOf'

Lukeisme:
in some browser, such as node webkit ,the event's path property is a nodeList object, so calling indexOf method will have an error.
using
```
 Array.prototype.indexOf.apply(Polymer.dom(event).path, [this])
```
instead


Lukeisme:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it!

PR #90, opened 2016-01-16
fix using 'indexOf' bug

Lukeisme:
in some browser, such as node webkit ,the event's path property is a nodeList object, so calling indexOf method will have an error.
using
```
Array.prototype.indexOf.apply(Polymer.dom(event).path, [this])
```
instead


namitos:
This as the previous code always returns -1. 
`event.path` not equal `Polymer.dom(event).path`. `this` element is not present in `event.path`, wrapped in `Polymer.dom()`
I don'k know why this event is wrapped

valdrinkoshi:
@Lukeisme can you please open an issue to explain the bug & provide an jsbin/example? Check the https://github.com/PolymerElements/iron-overlay-behavior/blob/master/CONTRIBUTING.md to get you started

PR #92, opened 2016-01-20
with-backdrop traps focus within overlay

### LINKED ISSUE #88: with-backdrop should trap the focus into the overlay
Currently when an overlay with backdrop is opened, it still allows the user to focus on elements behind the backdrop by using TAB
https://elements.polymer-project.org/elements/iron-overlay-behavior?view=demo:demo/index.html&active=Polymer.IronOverlayBehavior

Instead, the focus should be trapped within the overlay while opened
http://jsbin.com/koxoyu/3/edit?html,console,output

valdrinkoshi:
Fixes #88 by listening to focus changes on document.
`with-backdrop` will set the `tabindex` to `-1` so that focus can be trapped within the overlay.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -135,6 +135,18 @@ context. You should place this element as a child of `<body>` whenever possible.
         value: function() {
           return this._onCaptureKeydown.bind(this);
         }
+      },
+
+      _boundOnCaptureFocus: {
+        type: Function,
+        value: function() {
+          return this._onCaptureFocus.bind(this);
+        }
+      },
+
+      /** @type {?Node} */
+      _focusedChild: {
+        type: Object
       }
 
     },
@@ -152,10 +164,13 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     get _focusNode() {
-      return Polymer.dom(this).querySelector('[autofocus]') || this;
+      return this._focusedChild || Polymer.dom(this).querySelector('[autofocus]') || this;
     },
 
     ready: function() {
+      // with-backdrop need tabindex to be set in order to trap the focus.
+      // If it is not set, IronOverlayBehavior will set it, and remove it if with-backdrop = false.
+      this.__shouldRemoveTabIndex = false;
       this._ensureSetup();
     },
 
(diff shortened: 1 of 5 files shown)


iron-overlay-behavior.html
```diff
@@ -391,6 +428,18 @@
       }
     },
 
+    _onCaptureFocus: function (event) {
+      if (this._manager.currentOverlay() === this &&
+          this.withBackdrop) {
+        if (Polymer.dom(event).path.indexOf(this) === -1) {
+          event.stopPropagation();
```
valdrinkoshi: there is a specific line in the tests for this https://github.com/PolymerElements/iron-overlay-behavior/pull/92/files#diff-19c5ed5153f2ca6090daf300984fcad2R505

iron-overlay-behavior.html
```diff
@@ -54,6 +54,10 @@
 
   Polymer.IronOverlayBehaviorImpl = {
 
+    hostAttributes: {
```
valdrinkoshi: This is a breaking change, should remove it!

iron-overlay-behavior.html
```diff
@@ -265,6 +280,14 @@
...
 
     _withBackdropChanged: function() {
+      // If tabindex is already set, no need to override it.
+      if (this.withBackdrop && !this.hasAttribute('tabindex')) {
+        this.setAttribute('tabindex', '-1');
+        this.__shouldRemoveTabIndex = true;
+      } else if (this.__shouldRemoveTabIndex) {
+        this.removeAttribute('tabindex');
+        this.__shouldRemoveTabIndex = false;
+      }
```
cdata: Would it be sufficient to record the previous value of `tabindex` (if any) and then set it back if it is `!= null` when `withBackdrop` is `false`?

iron-overlay-behavior.html
```diff
@@ -391,6 +419,19 @@
       }
     },
 
+    _onCaptureFocus: function (event) {
+      if (this._manager.currentOverlay() === this &&
+          this.withBackdrop) {
+        var path = Polymer.dom(event).path;
+        if (path.indexOf(this) === -1) {
```
cdata: :beers:

iron-overlay-behavior.html
valdrinkoshi: Since we only need `tabindex` for focus purposes, any value is good. If someone wants to set a `tabindex=0` while `with-backdrop`, `iron-overlay-behavior` should allow it and not override it.

cdata: Word.

cdata:
LGTM :+1:

PR #93, opened 2016-01-26
call complete() if transitionDuration = 0s

### LINKED ISSUE #83: cannot interact with page if `iron-overlay-backdrop` transition is disabled
http://jsbin.com/qosiqix/1/edit?html,output

`iron-overlay-backdrop` will rely on the event `transitionend` to call the method `complete()` after it gets closed, which will handle the removal of the backdrop from the body if necessary.

### LINKED ISSUE #96: Backdrop not removed after call to dialog.close() on a modal dialog shortly after open
I am not sure I understand issue #83 and this might be the same, but I suspect not.  I also think it might have been introduced in a fairly recent change - because the code was working fairly recently. However I am not 100% sure of that as I have had to adjust timings of other things to handle a problem with theming. and it is possible that that change meant there are less things for the ajax request (see below) to queue behind.

It is a timing issue in that the dialog box that is opened and closed fairly rapidly. Let me try and explain ...

When I start up my application, I check for the existence of a cookie.  If it doesn't exist I put up a modal dialog box to collect user credentials.  The user clicks on an OK button it doesn't close the dialog box immediately, but rather sends an ajax request to the server to check credentials.  Only if it passes these credentials the server sets the cookie (with an authentication token) and the dialog closes.  This all works fine.

If the cookie exists, I briefly open a modal dialog box with a spinner in it.  I then send an ajax request to the server to ask it to validate the user. Assuming it does, the Ajax response causes me to close the modal dialog box again.

Until recently this all worked fine.  But over the past few days I have been experiencing a complete lock up of the user interface when the second case is true (ie we make this very quick Ajax call which causes me to open and then close the dialog rapidly).  Putting the debugger on things I have been able to see that the overlay is still in place after I have closed the dialog.  From the debugger, it reaches the close function of `iron-overlay-backdrop.html`, which attempts to set the `this.opened` flag to false.  The `_onTransitionend` function is not called.

The flight time of my ajax request seems to be about 6 milliseconds.

I have now added a 200 millisecond delay before closing the dialog and it started working again.

valdrinkoshi:
Fixes #83 by checking for css `transitionDuration` when closing the backdrop and calling `complete()` if duration is `0s`
Fixes #96 by checking of css `opacity` when closing the backdrop and calling `complete()` if `opacity` is already 0.
Updated previous test that was relying on disabling the animations.

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -119,9 +119,10 @@ Custom property | Description | Default
       // close only if no element with backdrop is left
       if (this._manager.getBackdrops().length === 0) {
         this._setOpened(false);
-        // complete() will be called after the transition is done.
-        // If animations are disabled via custom-styles, user is expected to call
-        // complete() after close()
+        // In case of no animations, complete
+        if (getComputedStyle(this).transitionDuration === '0s') {
+          this.complete();
+        }
       }
     },
 
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -118,9 +118,12 @@ Custom property | Description | Default
       this.style.zIndex = this._manager.backdropZ();
       // close only if no element with backdrop is left
       if (this._manager.getBackdrops().length === 0) {
+        // Read style before setting opened.
+        var cs = getComputedStyle(this);
+        var noAnimation = (cs.transitionDuration === '0s' || cs.opacity == 0);
         this._setOpened(false);
         // In case of no animations, complete
-        if (getComputedStyle(this).transitionDuration === '0s') {
+        if (noAnimation) {
           this.complete();
         }
       }
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -465,6 +465,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           overlays[0].opened = false;
           assert.equal(overlays[0]._manager.getBackdrops().length, 0, 'overlay removed from manager backdrops');
         });
+
       });
 
       suite('multiple overlays with backdrop', function() {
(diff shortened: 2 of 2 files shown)


iron-overlay-backdrop.html
```diff
@@ -118,10 +118,14 @@
       this.style.zIndex = this._manager.backdropZ();
       // close only if no element with backdrop is left
       if (this._manager.getBackdrops().length === 0) {
+        // Read style before setting opened.
+        var cs = getComputedStyle(this);
+        var noAnimation = (cs.transitionDuration === '0s' || cs.opacity == 0);
```
cdata: What is the extra condition about `opacity`? "Don't transition transparent things?"

valdrinkoshi: Exactly, because no transition will happen :)
Here we are in the `close()` method, which should go from `opacity: 1` to `opacity: 0`. If already 0, don't even bother waiting!

cdata: What if I want to transition from transparent to opaque as a close? :trollface:

cdata:
LGTM :+1: It might be good to add a test about that opacity condition, but I won't hold you to it.

iron-overlay-backdrop.html
valdrinkoshi: Here we don't wait if `opacity` is already 0, aka nothing's visible already :p
If that's not the case, we patiently wait for any animation to happen ;)
If you want to transition from transparent to opaque, this will do :grin: 
```html
<style is="custom-style">
  iron-backdrop-element {
    --iron-overlay-backdrop-background-color: transparent;
    --iron-overlay-backdrop: {
      transition: background-color 0.3s;
      opacity: 0.6;
    };
    --iron-overlay-backdrop-opened: {
      background-color: black;
    };
  }
</style>
```

valdrinkoshi:
I've already added the tests for that. Without the fix they fail (tested it ;) )

PR #98, opened 2016-02-03
add no-auto-focus test

valdrinkoshi:
Added tests for `no-auto-focus` property. Discovered "funky" behavior on Safari where if an element with `autofocus` attribute is displayed for the first time, it will receive the focus (tested with `button`, see example here http://jsbin.com/woroci/2/). So now `iron-overlay-behavior` handles also that case, making sure that the element is blurred synchronously after `overlay.open()` gets called.


valdrinkoshi:
@notwaldorf PTAL

notwaldorf:
I think this makes sense. LGTM

PR #104, opened 2016-02-11
test cleanup

valdrinkoshi:
Separate tests in their suites:
- basic overlay
- opened overlay
- focus handling
- overlay with backdrop
- multiple overlays
- z-ordering
- multiple overlays with backdrop
- a11y

Create utility methods to listen to events only once.
Use less instances of `test-fixture`

### DIFF
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -9,6 +9,7 @@ Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 <html>
+
   <head>
 
     <title>iron-overlay-behavior tests</title>
@@ -24,7 +25,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <link rel="import" href="test-overlay.html">
     <link rel="import" href="test-overlay2.html">
 
+    <style is="custom-style">
+      iron-overlay-backdrop {
+        /* For quicker tests */
+        --iron-overlay-backdrop: {
+          transition: none;
+        }
+      }
+    </style>
+
   </head>
+
   <body>
 
     <test-fixture id="basic">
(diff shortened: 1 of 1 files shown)


test/iron-overlay-behavior.html
```diff
@@ -52,55 +63,50 @@
...
-        <test-overlay class="overlay-3">
-          Overlay 3
+        <test-overlay with-backdrop>
+          Overlay with backdrop
         </test-overlay>
       </template>
     </test-fixture>
 
-    <test-fixture id="backdrop-multiple">
+    <test-fixture id="multiple">
```
notwaldorf: shouldn't `multiple` actually have multiple overlays? :)

test/iron-overlay-behavior.html
```diff
@@ -52,55 +63,50 @@
...
 
-      suite('basic overlay tests', function() {
+      function runAfterClose(overlay, callback) {
+        listenOnce(overlay, 'iron-overlay-closed', callback);
+        overlay.close();
+      }
+
+      function listenOnce(node, eventName, callback) {
+        var listener = function(event) {
+          node.removeEventListener(eventName, listener);
```
notwaldorf: why do we need to remove the listener?

test/iron-overlay-behavior.html
```diff
@@ -215,47 +191,21 @@
           });
         });
 
-        test('close an overlay in proximity to another overlay', function(done) {
```
notwaldorf: we don't need this test anymore?

test/iron-overlay-behavior.html
valdrinkoshi: will revert the runAfterOpen/Close

notwaldorf:
LGTM! Yay!

valdrinkoshi:
\o/

PR #106, opened 2016-02-11
Wrap focus in overlays with backdrop

### LINKED ISSUE #105: Focus trapping for modal dialogs doesn't wrap
_From @anthonytordillos on October 20, 2015 6:11_

Based on the WAI-ARIA best practices[1], it sounds like focus should wrap from the last focusable item in the dialog to first, and vice versa. As currently implemented, focus just stays on either the first or last focusable element depending on the direction of traversal. 

1. http://www.w3.org/TR/wai-aria-practices/#modal_dialog

_Copied from original issue: PolymerElements/paper-dialog-behavior#33_

valdrinkoshi:
Fixes #105 
We now wrap the focus for overlays `with-backdrop`. Respects `tabindex` too! :tada: 
Moreover, we also focus the first focusable element if there is no `[autofocus]` element.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -9,6 +9,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 -->
 
 <link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../iron-a11y-keys-behavior/iron-a11y-keys-behavior.html">
 <link rel="import" href="../iron-fit-behavior/iron-fit-behavior.html">
 <link rel="import" href="../iron-resizable-behavior/iron-resizable-behavior.html">
 <link rel="import" href="iron-overlay-backdrop.html">
@@ -118,6 +119,15 @@ context. You should place this element as a child of `<body>` whenever possible.
         type: Object
       },
 
+      /**
+       * The HTMLElement that will be firing relevant KeyboardEvents.
+       * Used for capturing esc and tab. Overridden from `IronA11yKeysBehavior`.
+       */
+      keyEventTarget: {
+        type: Object,
+        value: document
+      },
+
       _manager: {
         type: Object,
         value: Polymer.IronOverlayManager
@@ -130,13 +140,6 @@ context. You should place this element as a child of `<body>` whenever possible.
         }
       },
 
-      _boundOnCaptureKeydown: {
-        type: Function,
-        value: function() {
-          return this._onCaptureKeydown.bind(this);
-        }
-      },
-
       _boundOnCaptureFocus: {
         type: Function,
         value: function() {
@@ -144,11 +147,19 @@ context. You should place this element as a child of `<body>` whenever possible.
         }
       },
 
-      /** @type {?Node} */
+      /**
+       * The node being focused.
+       * @type {?Node}
+       */
       _focusedChild: {
         type: Object
       }
 
+    },
+
+    keyBindings: {
+      'esc': '__onEsc',
+      'tab': '__onTab'
     },
 
     listeners: {
@@ -157,20 +168,78 @@ context. You should place this element as a child of `<body>` whenever possible.
 
     /**
      * The backdrop element.
-     * @type Node
+     * @type {Node}
      */
     get backdropElement() {
       return this._manager.backdropElement;
     },
 
+    /**
+     * Returns the node to give focus to.
+     * @type {Node}
+     */
     get _focusNode() {
-      return this._focusedChild || Polymer.dom(this).querySelector('[autofocus]') || this;
+      return this._focusedChild || Polymer.dom(this).querySelector('[autofocus]') || this.__firstFocusableNode || this;
+    },
+
+    /**
+     * Array of nodes that can receive focus (overlay included), ordered by `tabindex`.
+     * This is used to retrieve which is the first and last focusable nodes in order
+     * to wrap the focus for overlays `with-backdrop`.
+     *
+     * If you know what is your content (specifically the first and last focusable children),
+     * you can override this method to return only `[firstFocusable, lastFocusable];`
+     * @type {[Node]}
+     * @protected
+     */
+    get _focusableNodes() {
+      // Elements that can be focused even if they have [disabled] attribute.
+      var FOCUSABLE_WITH_DISABLED = [
+        'a[href]',
+        'area[href]',
+        'iframe',
+        '[tabindex]',
+        '[contentEditable=true]'
+      ];
+
+      // Elements that cannot be focused if they have [disabled] attribute.
+      var FOCUSABLE_WITHOUT_DISABLED = [
+        'input',
+        'select',
+        'textarea',
+        'button'
+      ];
+
+      // Discard elements with tabindex=-1 (makes them not focusable).
+      var selector = FOCUSABLE_WITH_DISABLED.join(':not([tabindex="-1"]),') +
+        ':not([tabindex="-1"]),' +
+        FOCUSABLE_WITHOUT_DISABLED.join(':not([disabled]):not([tabindex="-1"]),') +
+        ':not([disabled]):not([tabindex="-1"])';
+
+      var focusables = Polymer.dom(this).querySelectorAll(selector);
+      if (this.tabIndex >= 0) {
+        // Insert at the beginning because we might have all elements with tabIndex = 0,
+        // and the overlay should be the first of the list.
+        focusables.splice(0, 0, this);
+      }
+      // Sort by tabindex.
+      return focusables.sort(function (a, b) {
+        if (a.tabIndex === b.tabIndex) {
+          return 0;
+        }
+        if (a.tabIndex === 0 || a.tabIndex > b.tabIndex) {
+          return 1;
+        }
+        return -1;
+      });
     },
 
     ready: function() {
       // with-backdrop need tabindex to be set in order to trap the focus.
       // If it is not set, IronOverlayBehavior will set it, and remove it if with-backdrop = false.
       this.__shouldRemoveTabIndex = false;
+      // Used for wrapping the focus on TAB / Shift+TAB.
+      this.__firstFocusableNode = this.__lastFocusableNode = null;
       this._ensureSetup();
     },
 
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
@notwaldorf @cdata kindly review

valdrinkoshi:
@cdata, your feedback is more than welcome too :)

iron-overlay-behavior.html
```diff
@@ -167,10 +167,33 @@
       return this._focusedChild || Polymer.dom(this).querySelector('[autofocus]') || this;
     },
 
+    /**
+     * Nodes that can receive focus (overlay included). They will be ordered by their tabindex.
+     * @type [Node]
+     */
+    get _focusableNodes() {
+      var focusables = Polymer.dom(this).querySelectorAll('a[href]:not([tabindex="-1"]),area[href]:not([tabindex="-1"]),input:not([disabled]):not([tabindex="-1"]),select:not([disabled]):not([tabindex="-1"]),textarea:not([disabled]):not([tabindex="-1"]),button:not([disabled]):not([tabindex="-1"]),iframe:not([tabindex="-1"]),[tabindex]:not([tabindex="-1"]),[contentEditable=true]:not([tabindex="-1"])');
```
notwaldorf: This looks...brittle? What's `area[href]`? Does this mean I won't get a `paper-button`?

iron-overlay-behavior.html
```diff
@@ -413,16 +444,28 @@
 
     _onCaptureKeydown: function(event) {
       var ESC = 27;
-      if (this._manager.currentOverlay() === this &&
-          !this.noCancelOnEscKey &&
-          event.keyCode === ESC) {
-        this.cancel(event);
+      var TAB = 9;
```
notwaldorf: Should this maybe use `a11yKeysBehavior`?

iron-overlay-behavior.html
valdrinkoshi: `paper-button` has `tabindex=0` which is selected via `[tabindex]:not([tabindex="-1"])` (towards the end of this line)

iron-overlay-behavior.html
valdrinkoshi: Makes sense, will look into it

iron-overlay-behavior.html
cdata: It seems like this string could be documented a little more cleanly by splitting out the elements into groups (as arrays) and generating this nasty selector string with code. Example:

```javascript
var AUTOFOCUSABLE_WITH_DISABLED = [
  'a',
  'area',
  'input',
  // etc...
]

var AUTOFOCUSABLE_WITHOUT_DISABLED = [
  '[tabindex]',
  'iframe',
  '[contenteditable=true]',
  // etc..
];
```

Then, iterate over the lists to generate the selector at registration time. This will enable the scary selector to be somewhat self-documenting. WDYT?

valdrinkoshi: good idea, will clean this up

test/iron-overlay-behavior.html
```diff
@@ -98,6 +125,31 @@
...
+          callback(event);
+        };
+        node.addEventListener(eventName, listener);
+      }
+
+      function pressTabAndChangeFocus(useShift, nodeToFocus) {
+        var TAB = 9;
+        if (useShift) {
+          // Limitation of MockInteractions for switching focus
+          var event = new CustomEvent('keydown', {
```
valdrinkoshi: Addressed in https://github.com/PolymerElements/iron-test-helpers/pull/45
Will refactor this once `modifiers` will be available

valdrinkoshi:
@cdata @notwaldorf feedback incorporated, PTAL :eyes:

iron-overlay-behavior.html
```diff
@@ -144,33 +146,91 @@
...
+    },
+
+    /**
+     * Array of nodes that can receive focus (overlay included), ordered by tabindex.
+     * @type {[Node]}
+     */
+    get _focusableNodes() {
+      // Elements that can be focused even if they have [disabled] attribute.
+      var FOCUSABLE_WITH_DISABLED = [
+        'a[href]',
```
notwaldorf: lol really? amazing. the web: where words don't have meanings.

iron-overlay-behavior.html
```diff
@@ -144,33 +146,91 @@
...
+      // Elements that cannot be focused if they have [disabled] attribute.
+      var FOCUSABLE_WITHOUT_DISABLED = [
+        'input',
+        'select',
+        'textarea',
+        'button'
+      ];
+
+      // Discard elements with tabindex=-1 (makes them not focusable).
+      var selector = FOCUSABLE_WITH_DISABLED.join(':not([tabindex="-1"]),') + ':not([tabindex="-1"]),' +
```
notwaldorf: would you mind breaking each of the `+` lines on a new line? Just so you can see all the components that go in this easier?

iron-overlay-behavior.html
```diff
@@ -239,7 +301,6 @@
         this.removeAttribute('aria-hidden');
       } else {
         this.setAttribute('aria-hidden', 'true');
-        Polymer.dom(this).unobserveNodes(this._observer);
```
notwaldorf: Isn't this being done in a different PR?

valdrinkoshi: yep you're right.

test/iron-overlay-behavior.html
```diff
@@ -98,6 +125,31 @@
         overlay.close();
       }
 
+      function listenOnce(node, eventName, callback) {
```
notwaldorf: Hahah, I like how `listenOnce` comes back in every PR 😉

test/iron-overlay-behavior.html
notwaldorf: 👍

test/iron-overlay-behavior.html
```diff
@@ -133,14 +185,23 @@
           assert.isFalse(spy.called, 'overlay should not refit');
         });
 
-        test('open() triggers iron-resize', function(done) {
+        test('open() & close() trigger iron-resize', function(done) {
```
notwaldorf: I think this was refactored in https://github.com/PolymerElements/iron-overlay-behavior/pull/108 as well right?

notwaldorf:
I'll hold off until https://github.com/PolymerElements/iron-overlay-behavior/pull/108 is merged, since a bunch of changes from here are in that PR.

test/iron-overlay-behavior.html
valdrinkoshi: lol i'll kill it with fire!

valdrinkoshi:
Ok, feedback incorporated (cosmetics too)!
The main problem is that `iron-overlay-behavior` won't be able to track children that are in other shadow roots (`querySelectorAll` will stops at the overlay's root). This is the main blocker for this PR. I could probably implement my own query selector but i'm afraid it will be more expensive in terms of performance. I'll give it a shot.

A less intrusive approach could be to exploit the user interaction (TAB & Shift+TAB) to find out which ones are the first focusable and last focusable. The limitation would be that if you open an overlay and then press `Shift + TAB`, it won't wrap (since we didn't find yet the last focusable element).

LMKWDYT :grin:

valdrinkoshi:
Ok, so I think this is the best it can get: now we don't rely on `_focusableNodes` to keep track of what's focusable, but we use it just to know what's the first & last focusable nodes.
Whoever uses `iron-overlay-behavior` could make `_focusableNodes` much more performant if they know what's their content (updated method's docs accordingly).

PTAL :pray:

iron-overlay-behavior.html
```diff
@@ -118,6 +119,14 @@
         type: Object
       },
 
+      /**
+       * The HTMLElement that will be firing relevant KeyboardEvents.
+       */
+      keyEventTarget: {
```
notwaldorf: Is this used anywhere? Or is it just to override the `IronA11yKeysBehavior` default? If the latter, maybe add that in the comment ("overriden from x") -- otherwise a future keen party might delete it by accident 😂

notwaldorf: Or even better, could you just set it in `ready` or something?

iron-overlay-behavior.html
```diff
@@ -462,6 +519,37 @@
       if (this.opened) {
         this.notifyResize();
       }
+      // Store it so we don't query too much.
+      var focusableNodes = this._focusableNodes;
+      this.__firstFocusableNode = focusableNodes[0];
+      this.__lastFocusableNode = focusableNodes[focusableNodes.length - 1];
```
notwaldorf: 👍

iron-overlay-behavior.html
```diff
@@ -462,6 +519,37 @@
...
+    },
+
+    __onTab: function(event) {
+      // Not opened or not on top, so return.
+      if (this._manager.currentOverlay() !== this) {
+        return;
+      }
+      // TAB wraps from last to first focusable.
+      // Shift + TAB wraps from first to last focusable.
+      var shift = event.detail.keyboardEvent.shiftKey;
```
notwaldorf: alternatively you could add a `keyBinding` for `shift + tab`, call `__onSfiftTab` and they both call a helper, but it's probably not worth it.

valdrinkoshi: both `tab` and `shift+tab` would get called, and I would have to do `event.preventDefault()` to avoid it :x

notwaldorf: 👍

test/iron-overlay-behavior.html
```diff
@@ -391,6 +418,86 @@
...
+          var focusableNodes = overlay._focusableNodes;
+          runAfterOpen(overlay, function() {
+            // Go to last element.
+            MockInteractions.focus(focusableNodes[focusableNodes.length-1]);
+            // Simulate TAB & focus out of overlay.
+            MockInteractions.pressAndReleaseKeyOn(document, 9);
+            MockInteractions.focus(document.body);
+            assert.equal(focusableNodes[0], document.activeElement, 'focus wrapped to first focusable');
+            // Simulate Shift+TAB & focus out of overlay.
+            MockInteractions.pressAndReleaseKeyOn(document, 9, ['shift']);
```
notwaldorf: Sweet!

notwaldorf:
LGTM (% I left a question about `keyTarget`)

iron-overlay-behavior.html
valdrinkoshi: I kind of liked the fact to set it here because if i set it on `ready` i would trigger observers. I'll update the doc mentioning `overridden from IronA11yKeysBehavior` ;)

PR #108, opened 2016-02-17
a11y & prettify demo page

### LINKED ISSUE #49: Mobile A11y Testing 11/1
ON Android with Talkback, upon opening these overlays, I receive no spoken feedback. This should speak out the title of the dialog as a spoken alert. Without seeing this visually, I have no idea that something is on the screen. 

Also, once I open the Hello World overlay, I have no clear way to get out of it.

Once an overlay opens, like the scrolling one, it should steal focus. Right now, if I open it and continue swiping linearly, I just go to the next buttons and don't have access to the scrolling content.

### LINKED ISSUE #97: iron-overlay-opened should be called after iron-resize
Currently, `iron-resize` gets called after `iron-overlay-opened` http://jsbin.com/xeyeqef/1/edit?html,console,output
Shouldn't `iron-overlay-opened` be called only after `iron-resize` instead?

valdrinkoshi:
Fixes #49, fixes #97 by setting `tabindex=-1` to all overlays so they can be focusable and ensuring `iron-resize` is called before `iron-overlay-opened`. Moreover we now call `notifyResize()` before `_applyFocus()` so that we are sure `refit()` will make the overlay visible (setting `display`).
Additionally, we attach the node change observer only once.
The demo page now looks like this:
overlay-pretty-demo

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -179,9 +179,12 @@ context. You should place this element as a child of `<body>` whenever possible.
       if (this._callOpenedWhenReady) {
         this._openedChanged();
       }
+      this._observer = Polymer.dom(this).observeNodes(this._onNodesChange);
     },
 
     detached: function() {
+      Polymer.dom(this).unobserveNodes(this._observer);
+      this._observer = null;
       this.opened = false;
       this._manager.trackBackdrop(this);
       this._manager.removeOverlay(this);
@@ -212,7 +215,7 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     /**
-     * Cancels the overlay. 
+     * Cancels the overlay.
      * @param {?Event} event The original event
      */
     cancel: function(event) {
@@ -239,7 +242,6 @@ context. You should place this element as a child of `<body>` whenever possible.
         this.removeAttribute('aria-hidden');
       } else {
         this.setAttribute('aria-hidden', 'true');
-        Polymer.dom(this).unobserveNodes(this._observer);
       }
 
       // wait to call after ready only if we're initially open
@@ -330,6 +332,8 @@ context. You should place this element as a child of `<body>` whenever possible.
 
       this._manager.addOverlay(this);
 
+      // Needed to calculate the size of the overlay so that transitions on its size
+      // will have the correct starting points.
       this._preparePositioning();
       this.fit();
       this._finishPositioning();
@@ -362,23 +366,24 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     _finishRenderOpened: function() {
-      // focus the child node with [autofocus]
+      // This ensures the overlay is visible before we set the focus
+      // (by calling _onIronResize -> refit).
+      this.notifyResize();
+      // Focus the child node with [autofocus]
       this._applyFocus();
 
-      this._observer = Polymer.dom(this).observeNodes(this.notifyResize);
       this.fire('iron-overlay-opened');
     },
 
     _finishRenderClosed: function() {
-      // hide the overlay and remove the backdrop
+      // Hide the overlay and remove the backdrop.
       this.resetFit();
       this.style.display = 'none';
       this._manager.removeOverlay(this);
 
-      this._focusedChild = null;
       this._applyFocus();
-
       this.notifyResize();
+
       this.fire('iron-overlay-closed', this.closingReason);
     },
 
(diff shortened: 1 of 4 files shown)


valdrinkoshi:
@notwaldorf @cdata FYI, kindly review

iron-overlay-behavior.html
```diff
@@ -330,6 +332,8 @@
 
       this._manager.addOverlay(this);
 
+      // Needed to calculate the size of the overlay so that transitions on its size
+      // will have the correct starting points.
```
notwaldorf: 💯

iron-overlay-behavior.html
```diff
@@ -391,8 +396,9 @@
     _finishPositioning: function() {
       this.style.display = 'none';
       this.style.transform = this.style.webkitTransform = '';
-      // force layout to avoid application of transform
-      /** @suppress {suspiciousCode} */ this.offsetWidth;
+      // Force layout layout to avoid application of transform.
+      // Set offsetWidth to itself so that compilers won't remove it.
+      this.offsetWidth = this.offsetWidth;
```
notwaldorf: Are you sure this doesn't perform a style recalc twice (once to get it, once to set it?)

notwaldorf: Ok just checked with Levi and it's fine (you only do a style recalc if you need to, and after the first `offsetWidth` the second will be the same). But are you sure we shouldn't leave `@suppress {suspiciousCode}` in, in case the compiler still wants to remove it as a no-op?

test/iron-overlay-behavior.html
```diff
@@ -133,14 +141,23 @@
...
+        test('open() & close() trigger iron-resize', function(done) {
           // Ignore iron-resize triggered by window resize.
           var callCount = 0;
           window.addEventListener('resize', function() { callCount--; }, true);
           overlay.addEventListener('iron-resize', function() { callCount++; });
-          runAfterOpen(overlay, function() {
-            assert.equal(callCount, 1, 'iron-resize should be called once');
-            done();
+          overlay.open();
+          listenOnce(overlay, 'iron-overlay-opened', function() {
```
notwaldorf: listeners all the way downnn. 🐢 

Would this be clearer if it was just 2 tests? `open() triggers`, `close() triggers`

iron-overlay-behavior.html
valdrinkoshi: Compilers should not remove this line at all, since it is the one that enables/forces the animations.

test/iron-overlay-behavior.html
valdrinkoshi: That would make the `close() triggers` test a replica of `open() triggers` test + another close. That's why I ended up merging these two.
But probably is better to separate these 2. I'll do it

iron-overlay-behavior.html
notwaldorf: I don't know `a = a` is a exactly the kind a compiler might optimize away (much like how `a;` was optimized away). My understanding was  the compiler doesn't know what `this.offsetWidth` does, and that's why `@suppress {suspiciousCode}` was needed. Maybe ping @rictic, our human compiler, for a double check?

test/iron-overlay-behavior.html
```diff
@@ -139,11 +139,24 @@
...
         });
 
+        test('close() triggers iron-resize', function(done) {
+          // Ignore iron-resize triggered by window resize.
+          var callCount = 0;
+          window.addEventListener('resize', function() { callCount--; }, true);
+          overlay.addEventListener('iron-resize', function() { callCount++; });
+          runAfterOpen(overlay, function() {
+            runAfterClose(overlay, function() {
+              assert.equal(callCount, 3, 'iron-resize called once before iron-overlay-closed');
```
notwaldorf: 3 != "once" -- does the comment need to be updated? :)

notwaldorf:
LGTM with a nit, but pretty please double check with @rictic about the `@suppress {suspiciousCode}` bit

test/iron-overlay-behavior.html
valdrinkoshi: I'll update it to be `assert.equal(callCount - 2, 1` ;)

valdrinkoshi:
@notwaldorf feedback incorporated, now tests are much cleaner!
Regarding the `@suppress {suspiciousCode}`, please see this related [issue on `iron-collapse`](https://github.com/PolymerElements/iron-collapse/issues/34) and [its fix](https://github.com/PolymerElements/iron-collapse/pull/35)

PR #110, opened 2016-02-22
update events docs

### LINKED ISSUE #103: iron-overlay-closed event doc is garbled
https://elements.polymer-project.org/elements/iron-overlay-behavior#event-iron-overlay-closed

    * @param {{canceled: (boolean|undefined)}} set to the `closingReason` attribute

Not sure exactly what this is supposed to look like. The description isn't really helpful.

I'm not sure whether the @param tag to an event should assume that the parameters are all passed in the detail object, or whether they may be added directly to the event object. (Maybe @garlicnation can help out here?)

In the first case, this should probably be:

    * @param {boolean|undefined} cancelled Whether the overlay was cancelled.

In the latter case, I think it should be something like:

    * @param {{canceled: (boolean|undefined)}} detail Set to the `closingReason` object.

Also, we really need to work on formatting those closure type definitions better in the output. They are fugly.

valdrinkoshi:
Fixes #103, final result is:
screen shot 2016-02-22 at 11 19 04 am

Also, moved docs at the bottom otherwise we would have duplicates (see https://elements.polymer-project.org/elements/iron-overlay-behavior#event-iron-overlay-canceled)

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -552,28 +552,29 @@ context. You should place this element as a child of `<body>` whenever possible.
         this._focusedChild = nodeToSet;
       }
     }
-
-/**
- * Fired after the `iron-overlay` opens.
- * @event iron-overlay-opened
- */
-
-/**
- * Fired when the `iron-overlay` is canceled, but before it is closed.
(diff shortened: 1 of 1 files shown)


notwaldorf:
LGTM

PR #111, opened 2016-02-22
remove autofocus of first element

valdrinkoshi:
revert the auto focusing of the first focusable element, since it is a breaking change (introduced in https://github.com/PolymerElements/iron-overlay-behavior/pull/106) that no-one asked.


notwaldorf:
LGTM

PR #112, opened 2016-02-23
events should not bubble

### LINKED ISSUE #70: using paper-dropdown-menu inside a dialog will close it
it seems that there is another problem with the https://github.com/PolymerElements/iron-overlay-behavior/pull/64.

I've made you a JSBin

https://jsbin.com/lapedorogo/edit?html,console,output

I don't know why, but if you use paper-dropdown-menu + paper-menu + paper-item inside a dialog (dialog and paper-dropdown-menu are differents module), when selecting an item in the paper-dropdown-menu, an event will be fired which seems to be "iron-overlay-closed"

Update : no need to select an item, clicking outside to hide the list will have the same effect

valdrinkoshi:
Fixes #70 by setting `bubbles: false` to the events.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -288,7 +288,7 @@ context. You should place this element as a child of `<body>` whenever possible.
      * @param {?Event} event The original event
      */
     cancel: function(event) {
-      var cancelEvent = this.fire('iron-overlay-canceled', event, {cancelable: true});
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
@cdata FYI, do you foresee any issues with this change? I could not.

notwaldorf:
Isn't this totally a breaking change? :(

valdrinkoshi:
Yep, updated the issue as well. Closing this for now, might be useful in the future though.

PR #114, opened 2016-02-24
call notifyResize only when not animating. Call refit only once on opened changes.

### LINKED ISSUE #113: prepareRenderOpened should call refit, not fit
Currently `iron-overlay-behavior` calls `fit` on opened (`_prepareRenderOpened`), and `resetFit` on close (`_finishRenderClosed`).
Instead, it should only call `refit` on `_prepareRenderOpened`, since the measuring of the dimensions matters in that moment, not when the overlay gets closed.

valdrinkoshi:
Fixes #113 by calling `refit` on `_prepareRenderedOpened`. Moreover limit the calls to `notifyResize` when the animation is ongoing; also in `_onIronResize` callback, we don't call `refit` if the animation is not done yet. Updated tests accordingly.

This helps fixing issues like [this one on `iron-dropdown`](https://github.com/PolymerElements/iron-dropdown/pull/63#issuecomment-187840348): what happens there is that `_prepareRenderOpened` calls only `fit`, so it won't recalculate the position information when `opened` changes. This PR makes sure we call `refit` right when (and only when) it is needed.

### DIFF
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -165,8 +165,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         });
 
         test('open() triggers iron-resize', function(done) {
-          // Ignore the iron-resize on attached.
-          var callCount = -1;
+          var callCount = 0;
           // Ignore iron-resize triggered by window resize.
           window.addEventListener('resize', function() { callCount--; }, true);
           overlay.addEventListener('iron-resize', function() { callCount++; });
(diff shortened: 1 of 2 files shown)


test/iron-overlay-behavior.html
```diff
@@ -165,8 +165,7 @@
         });
 
         test('open() triggers iron-resize', function(done) {
-          // Ignore the iron-resize on attached.
-          var callCount = -1;
+          var callCount = 0;
```
valdrinkoshi: So that first `iron-resize` was caused by the node observer which is added on attached. Now that we make sure it is not called during animations, this call count can finally be 0!

notwaldorf:
LGTM

PR #115, opened 2016-02-24
allow 1px of delta when checking position

valdrinkoshi:
Fixes tests failing on master (most likely introduced by the latest version of `iron-fit-behavior`). We now allow 1px of delta when checking the position of overlays.


notwaldorf:
LGTM

PR #118, opened 2016-02-26
no string literals in IronOverlayManager

valdrinkoshi:
The manager currently uses string literals to define getters for certain properties. This is a problem when compilers rename the properties, because string literals won't be modified.
To solve this, rather than using `Object.defineProperty` (which requires a string for the property name), we return an object and define the properties like:
```javascript
var o = {
  get prop(){
    return 'foo';
  }
```

Moreover, we now use an IIFE; this allows not to expose the class, and to make private properties methods really private/contained in the scope of the IIFE.

This PR makes it impossible to access to private objects (e.g. `Polymer.IronOverlayManager._overlays`), but all these were meant to be private so this should not be considered a breaking change.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -136,6 +136,11 @@ context. You should place this element as a child of `<body>` whenever possible.
         value: false
       },
 
+      /**
+       * Shortcut to access to the overlay manager.
+       * @private
+       * @type {Polymer.IronOverlayManagerClass}
+       */
       _manager: {
         type: Object,
         value: Polymer.IronOverlayManager
@@ -176,7 +181,7 @@ context. You should place this element as a child of `<body>` whenever possible.
 
     /**
      * The backdrop element.
-     * @type {Node}
+     * @type {Element}
      */
     get backdropElement() {
       return this._manager.backdropElement;
@@ -197,7 +202,7 @@ context. You should place this element as a child of `<body>` whenever possible.
      *
      * If you know what is your content (specifically the first and last focusable children),
      * you can override this method to return only `[firstFocusable, lastFocusable];`
-     * @type {[Node]}
+     * @type {Array<Node>}
      * @protected
      */
     get _focusableNodes() {
@@ -295,7 +300,7 @@ context. You should place this element as a child of `<body>` whenever possible.
 
     /**
      * Cancels the overlay.
-     * @param {?Event} event The original event
+     * @param {Event=} event The original event
      */
     cancel: function(event) {
       var cancelEvent = this.fire('iron-overlay-canceled', event, {cancelable: true});
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -15,197 +15,244 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   /**
    * @struct
    * @constructor
+   * @private
    */
   Polymer.IronOverlayManagerClass = function() {
+    /**
+     * Used to keep track of the opened overlays.
+     * @private {Array<Element>}
+     */
     this._overlays = [];
-    // Used to keep track of the last focused node before an overlay gets opened.
+
+    /**
+     * Used to keep track of the last focused node before an overlay gets opened.
+     * @private {Array<Element>}
+     */
     this._lastFocusedNodes = [];
 
     /**
-     * iframes have a default z-index of 100, so this default should be at least
-     * that.
+     * iframes have a default z-index of 100,
+     * so this default should be at least that.
      * @private {number}
      */
     this._minimumZ = 101;
 
+    /**
+     * Used to keep track of the opened overlays with backdrop.
+     * @private {Array<Element>}
+     */
     this._backdrops = [];
 
+    /**
+     * Memoized backdrop element.
+     * @private {Element|null}
+     */
     this._backdropElement = null;
-    Object.defineProperty(this, 'backdropElement', {
-      get: function() {
-        if (!this._backdropElement) {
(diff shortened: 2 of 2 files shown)


valdrinkoshi:
@rictic @cdata kindly review.

rictic:
Did you make any other changes as part of this? I notice that you've moved the private methods into the closure rather than naming them with underscores.

valdrinkoshi:
@notwaldorf FYI (LMKWDYT about code style & if you consider it a breaking change)

rictic:
Have you confirmed that this compiles cleanly?

valdrinkoshi:
@rictic tested, corrected some jsdoc things, now it builds cleanly :tada: 
sponge/c9c3562b-a7a2-4dd9-bd6b-8efe537758b4

valdrinkoshi:
@rictic, probably I could also just return an object like this:
```javascript
Polymer.IronOverlayManager = {
  _overlays: [],
// ...
};
```
I see this was already the case before #75, where you converted the manager in a singleton class, but I can't recall the reason behind that change.
@justinfagnani FYI

valdrinkoshi:
Alright, just found out that the compiler would consider it dangerous to use `this` internally (referring to a global object)
https://sponge.corp.google.com/invocation?tab=Build+Log&id=9b2db16a-e7fb-4f50-adb2-6e900b6fd86c
So either we use `Polymer.IronOverlayManager` everywhere instead of `this` (ugly), or use an IIFE.
WDYT?

iron-overlay-behavior.html
```diff
@@ -293,7 +293,7 @@
 
     /**
      * Cancels the overlay.
-     * @param {?Event} event The original event
+     * @param {Event} [event] The original event
```
rictic: What's with the []s here?

valdrinkoshi: `event` is an optional param, so it was a typo

valdrinkoshi:
The 4 commits are the possible 4 versions to do the same thing.
[Manager as object](https://github.com/PolymerElements/iron-overlay-behavior/commit/882689a7c926f47c71e25c9614c8c9344a6d6f9e) (my favorite) would fail because of the usage of `this`, so the next favorite would be [with IFEE and var](https://github.com/PolymerElements/iron-overlay-behavior/commit/47117f5c24be7e974d495d745d994a92ea1b2d50)

iron-overlay-manager.html
```diff
@@ -11,201 +11,186 @@
...
-   * @returns {Boolean}
-   */
-  Polymer.IronOverlayManagerClass.prototype._isChildOfOverlay = function(node) {
-    while (node && node !== document.body) {
-      // Use logical parentNode, or native ShadowRoot host.
-      node = Polymer.dom(node).parentNode || node.host;
-      // Check if it is an overlay.
-      if (node && node.behaviors && node.behaviors.indexOf(Polymer.IronOverlayBehaviorImpl) !== -1) {
-        return true;
+    get deepActiveElement() {
```
justinfagnani: if getters are in all our supported browsers, then why not just change the defineProperties to getters? Much less of a change and gets rid of the literals.

valdrinkoshi:
Alright, so the proper using of `@this` and `@typedef` allowed to go with the option Manager as an object :) Compiler builds it cleanly http://sponge/3a827a04-7346-485e-beaf-780fd5700134
@rictic kindly review

iron-overlay-behavior.html
rictic: Does `[event]` mean that `event` is an optional param?

I don't see anything about that syntax here: https://developers.google.com/closure/compiler/docs/js-for-compiler

iron-overlay-manager.html
```diff
@@ -13,199 +13,261 @@
 <script>
 
   /**
-   * @struct
-   * @constructor
+   * Singleton instance.
+   * This is compiler-friendly, because properties can be safely renamed.
+   * @typedef {Object} Polymer.IronOverlayManager
```
rictic: Huh, I didn't realize this would work. This does type checking? If you misspell a property of IronOverlayManager, or treat a property as the wrong type, does the compiler throw an error?

If not, that's great!

rictic:
Two small comments, otherwise this LGTM

iron-overlay-behavior.html
valdrinkoshi: You're right, I should make this optional with the notation `{Event=} event`

iron-overlay-manager.html
valdrinkoshi: Unfortunately it doesn't do any of the above. The error is thrown if we misspell a function name, but that's it :/

iron-overlay-manager.html
```diff
@@ -15,197 +15,244 @@
...
-    //
-    // NOTE: We make the assumption that any overlay that completes a transition
-    // will call into focusOverlay to kick the process back off. Currently:
-    // transitionend -> _applyFocus -> focusOverlay.
-    if (current && !current.transitioning) {
-      current._applyFocus();
-    }
-  };
+    focusOverlay: function() {
+      var current = /** @type {?} */ (this.currentOverlay());
```
valdrinkoshi: This `@type` is needed since we access to `transitioning` and `_applyFocus`, otherwise the compiler complains.

valdrinkoshi: @rictic ideally we should have a `typedef` for the elements that extend `IronOverlayBehavior`, but that should be defined in `iron-overlay-behavior`. I believe it's too much overhead, WDYT?

valdrinkoshi:
@rictic, ready for another review, PTAL :eyes:

rictic:
Looks good to me!

PR #120, opened 2016-03-07
Update iron-overlay-behavior.html

kysonic:
Sometimes auto refitting could be annoying, especially if overlay contains some dynamic animated elements like dropdown or context menu.


valdrinkoshi:
Hi @kysonic, thanks for the contribution!
Could you please first open an issue to discuss about the problem, so we can verify it is actually an issue? (see more details here https://github.com/PolymerElements/iron-overlay-behavior/blob/master/CONTRIBUTING.md)
Closing this for now

PR #122, opened 2016-03-08
check for truthy (isObject not reliable with Element)

valdrinkoshi:
Fixes tests on master, currently failing because of the new version of chai (the method `isObject` doesn't like anymore `Element`)


notwaldorf:
LGTM

PR #123, opened 2016-03-08
use requestAnimationFrame for non-blocking animations

### LINKED ISSUE #99: use requestAnimationFrame
It would improve performance to delay `_onIronResize` via `requestAnimationFrame`. e.g.
```javascript
    
    // properties...
    _boundOnIronResize: {
        type: Function,
        value: function() {
          return requestAnimationFrame(this._onIronResize.bind(this));
        }
      }
    //...
    listeners: {
      'iron-resize': '_boundOnIronResize'
    },
    //...
    _onIronResize: function(){
       //can be overridden
    }
```
Same for `_renderOpened/_renderClosed`

### LINKED ISSUE #135: Cannot set property offsetWidth of #<HTMLElement> which has only a getter
I get this error: 
Cannot set property offsetWidth of HTMLElement which has only a getter
In this line 
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L328

My Chrome version is 49, and here is some info - offsetWidth is read-only property
http://stackoverflow.com/questions/30453078/uncaught-typeerror-cannot-set-property-offsetwidth-of-htmlelement-which-has

One more proof that it's readonly property https://developer.mozilla.org/ru/docs/Web/API/HTMLElement/offsetWidth

valdrinkoshi:
Fixes #99, fixes #135 by using `requestAnimationFrame` instead of `async`. 
Also, removed unnecessary setting of `display: ''` followed by `display: none`, and then again `display: ''`. It is enough to just make the overlay visible, calculate its size, and then `requestAnimationFrame` will cause the relayout. 
Tested the patch on `paper-dialog` with animations, it doesn't flash in any of the tested browsers (chrome, firefox, safari).

Also, reset `z-index` only on `_finishRenderClosed`, so that the overlay is on top during its animation, and goes "behind" only when is invisible. This makes the dialog animations much better because a dialog with backdrop doesn't suddenly go back the backdrop when closed.

#135 gets fixed because we don't need to set `offsetWidth` to itself anymore.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -226,6 +226,10 @@ context. You should place this element as a child of `<body>` whenever possible.
       this.__shouldRemoveTabIndex = false;
       // Used for wrapping the focus on TAB / Shift+TAB.
       this.__firstFocusableNode = this.__lastFocusableNode = null;
+      // Used for requestAnimationFrame when opened changes.
+      this.__openChangedAsync = null;
+      // Used for requestAnimationFrame when iron-resize is fired.
+      this.__onIronResizeAsync = null;
       this._ensureSetup();
     },
 
@@ -313,23 +317,19 @@ context. You should place this element as a child of `<body>` whenever possible.
       if (this.opened) {
         this._prepareRenderOpened();
       }
-      if (this._openChangedAsync) {
-        this.cancelAsync(this._openChangedAsync);
+
+      // requestAnimationFrame for non-blocking rendering
+      if (this.__openChangedAsync) {
+        cancelAnimationFrame(this.__openChangedAsync);
       }
-      // Async here to allow overlay layer to become visible.
-      this._openChangedAsync = this.async(function() {
-        // overlay becomes visible here
-        this.style.display = '';
-        // Force layout to ensure transition will go. Set offsetWidth to itself
-        // so that compilers won't remove it.
-        this.offsetWidth = this.offsetWidth;
+      this.__openChangedAsync = requestAnimationFrame(function() {
+        this.__openChangedAsync = null;
         if (this.opened) {
           this._renderOpened();
         } else {
           this._renderClosed();
         }
-        this._openChangedAsync = null;
-      });
+      }.bind(this));
     },
 
     _canceledChanged: function() {
@@ -425,6 +425,8 @@ context. You should place this element as a child of `<body>` whenever possible.
     _finishRenderClosed: function() {
       // Hide the overlay and remove the backdrop.
       this.style.display = 'none';
+      // Reset z-index only at the end of the animation.
+      this.style.zIndex = '';
 
       this._applyFocus();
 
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
Safari 9 failure on travis is a lie, I've run the same tests from saucelabs and they pass https://saucelabs.com/beta/tests/08610528163449f1a586c1635f49d627/watch

iron-overlay-behavior.html
```diff
@@ -308,28 +312,25 @@
 
       this._manager.addOrRemoveOverlay(this);
 
+      // TODO handle case when it was already animating, e.g. quick open/close.
```
cdata: Include some signature next to your TODO, e.g., `// TODO(cdata):` - this way someone who reads this in the future can hunt you down and shame you :smiling_imp:

valdrinkoshi: lol makes sense. About this TODO in particular, it still needs more thinking hence it maybe doesn't make sense to add it at all.
e.g. if I `open()` then `close()`, what damage can do to trigger `iron-overlay-closed` if `iron-overlay-opened` wasn't triggered before it?
WDYT?

valdrinkoshi: I'll remove this TODO as I think it would be over engineering, & without a real use case it would make this behavior just slower for nothing.

iron-overlay-behavior.html
```diff
@@ -313,23 +317,19 @@
       if (this.opened) {
         this._prepareRenderOpened();
       }
-      if (this._openChangedAsync) {
-        this.cancelAsync(this._openChangedAsync);
+
+      // requestAnimationFrame for non-blocking rendering
+      if (this.__openChangedAsync) {
```
cdata: It might be good to call this `__openChangedRaf` since it's not generated by `this.async`..

iron-overlay-behavior.html
```diff
@@ -528,12 +526,15 @@
...
+        this.__onIronResizeAsync = null;
       }
-
-      if (this.opened) {
-        this.refit();
+      if (this.opened && !this.__isAnimating) {
+        this.__onIronResizeAsync = requestAnimationFrame(function() {
+          this.__onIronResizeAsync = null;
+          this.refit();
+        }.bind(this));
```
cdata: It would be kind of nice if we could just tell `debounce` to use `requestAnimationFrame`... (no action required).

cdata:
Minor naming issue regarding `async` vs `raf`. Feel free to make the call, it doesn't need to block this PR if you think it doesn't matter that much.

LGTM :+1:

valdrinkoshi:
About `async` vs `raf`, I think `async` is more communicative than `raf`, because it tells "this won't happen synchronously". With `raf` I'd have to search for the meaning of it, or look at the implementation :p
Also, is a private property, so I'll keep it like this for now :grin:

Issue #124, opened 2016-03-10
Wrong backdrop z-index for multiple overlays


r0l1:
If multiple dialogs are opened with a backdrop, then the backdrop should update its z-index to overlay all previous dialogs.

**Demonstration:**

Use the demo source in the demo/index.html file and adjust the "Multiple overlays can be opened" sample, by adding the _with-backdrop_ attribute to both overlays:

``` html
<h3>Multiple overlays can be opened.</h3>
<demo-snippet>
<template>
  <button onclick="multiple.open()">Open first overlay</button>
  <simple-overlay id="multiple" tabindex=-1 with-backdrop>
    <p>click to open another overlay</p>
    <button onclick="multiple2.open()">Open second overlay</button>
    <button onclick="multiple.close()">Close this</button>
  </simple-overlay>
  <simple-overlay id="multiple2" tabindex=-1 with-backdrop>
    <h2>Hi!</h2>
    <button onclick="multiple2.close()">Close</button>
  </simple-overlay>
</template>
</demo-snippet>
```

m4ng0squ4sh:
Hmm, I understand. However I do think, there should be an option available to set this specific behaviour. Imagine an open dialog displaying some data. Data manipulation is triggered by the user and a confirmation message box pops up. The backdrop overlaying the previous dialog is strictly required.Otherwise the user would be able to trigger further data manipulations on the dialog underneath, although the message box is still visible.

An additional attribute to trigger this behaviour would be fantastic:

```html
<simple-overlay with-backdrop attached-backdrop>
</simple-overlay>

<!-- OR -->

<simple-overlay with-backdrop sticky-backdrop>
</simple-overlay>

<!-- OR -->

<simple-overlay with-backdrop pin-backdrop>
</simple-overlay>
```

What do you think?

valdrinkoshi:
Uhm, there is another feature request for a new property `always-on-top` https://github.com/PolymerElements/iron-overlay-behavior/issues/3
Maybe that could be used together with `with-backdrop` to have the overlay with `always-on-top` to be on top of the others & the backdrop element too.

I wonder though if the backdrop should be always behind the top overlay by default.

m4ng0squ4sh:
Sounds good. Could you explain the functionality of #3 in detail? There is no description and I didn't understand the full context of `always-on-top`. I would then work on it and provide a pull request. I really need this feature soon and I don't want to hack around this problem...

[...]

PR #125, opened 2016-03-10
fixed bug #124

m4ng0squ4sh:


m4ng0squ4sh:
I signed it!

m4ng0squ4sh:
This change causes an issue if a toast is active while closing the overlay. The backdrop won't update it's z-index...

PR #126, opened 2016-03-11
always-on-top implementation

### LINKED ISSUE #3: Implement `always-on-top`
aka `layered`

valdrinkoshi:
Fixes #3 by implementing `always-on-top` property.
- the overlay with `always-on-top` should match the `Polymer.IronOverlayManager.currentOverlay()` even if other overlays are opened after it
- if there is more than one overlay with `always-on-top`, the last one to be opened wins (is on top)
- `with-backdrop` doesn't affect `always-on-top`

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -85,8 +85,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _bringOverlayAtIndexToFront: function(i) {
       var overlay = this._overlays[i];
       var lastI = this._overlays.length - 1;
+      // Ensure always-on-top overlay stays on top.
+      if (!overlay.alwaysOnTop && this._overlays[lastI].alwaysOnTop) {
+        lastI--;
+      }
       // If already the top element, return.
-      if (!overlay || i === lastI) {
+      if (!overlay || i >= lastI) {
         return;
       }
       // Update z-index to be on top.
@@ -119,6 +123,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     /**
      * Tracks overlays for z-index and focus management.
+     * Ensures the last added overlay with always-on-top remains on top.
      * @param {Element} overlay
      */
     addOverlay: function(overlay) {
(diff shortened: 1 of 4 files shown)


iron-overlay-manager.html
```diff
@@ -243,7 +267,7 @@
     _getOverlayZ: function(overlay) {
       var z = this._minimumZ;
       if (overlay) {
-        var z1 = Number(window.getComputedStyle(overlay).zIndex);
+        var z1 = Number(overlay.style.zIndex || window.getComputedStyle(overlay).zIndex);
```
notwaldorf: hmm when would `overlay.style.zIndex` != `window.getComputedStyle(overlay).zIndex` ?

iron-overlay-manager.html
```diff
@@ -132,6 +145,17 @@
         element = null;
       }
       this._lastFocusedNodes.push(element);
+
+      // Ensure always-on-top overlay stays on top.
+      if (onTop) {
+        minimumZ = this.currentOverlayZ();
+        this._overlays.push(onTop);
```
notwaldorf: I would add a comment that `onTop` has already been spliced out of `this._overlays`, because it happens 25 lines up and it's not obvious. Or even better, can you do the splicing here?

iron-overlay-manager.html
valdrinkoshi: it happens during the splicing in `addOverlay`. 
Also, using `overlay.style.zIndex` doesn't cause a relayout, so it should be more performant :grin:

iron-overlay-manager.html
valdrinkoshi: I agree, it can be written to be more clear :+1:

iron-overlay-manager.html
```diff
@@ -117,21 +117,38 @@
...
+        // Update minimumZ to match previous overlay's z-index.
+        var previousOverlay = this._overlays[insertionIndex - 1];
+        minimumZ = Math.max(this._getOverlayZ(previousOverlay), this._minimumZ);
+      }
+
+      // Update z-index and insert overlay.
       if (newZ <= minimumZ) {
         this._applyOverlayZ(overlay, minimumZ);
       }
+      this._overlays.splice(insertionIndex, 0, overlay);
```
notwaldorf: I don't see where the push into `this._overlays` is happening anymore, but I think i'm missing something

valdrinkoshi: [`splice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) can be used to remove + insert new elements, so all I do now is find the right insertion index and insert the new overlay in the right spot

notwaldorf: But wasn't the overlay going on top of everything, since it's the last one being added? Now you're inserting it into a different place than before (whether or not it's the "always on top" one)

valdrinkoshi: That would be 2 operations on the array. With my implementation instead we have only 1 operation on the array, that is insert the new overlay at its right place.

valdrinkoshi:
Tests failing because safari7, I've run them locally with this patch and they're green https://github.com/PolymerElements/test-fixture/pull/32

iron-overlay-manager.html
notwaldorf: Sure, but isn't it in the wrong spot? If before it was getting pushed on top, and now you're inserting it randomly in the array, isn't that in a different spot?

valdrinkoshi: No, the initial value of `insertionIndex` is `this._overlays.length` (aka on top). It gets decremented of 1 if the currentOverlay should be kept on top instead of overlay.

iron-overlay-manager.html
```diff
@@ -85,8 +85,12 @@
     _bringOverlayAtIndexToFront: function(i) {
       var overlay = this._overlays[i];
       var lastI = this._overlays.length - 1;
+      // Ensure always-on-top overlay stays on top.
+      if (!overlay.alwaysOnTop && this._overlays[lastI].alwaysOnTop) {
+        lastI--;
+      }
       // If already the top element, return.
-      if (!overlay || i === lastI) {
+      if (!overlay || i >= lastI) {
```
notwaldorf: Why isn't this needed anymore?

notwaldorf:
LGTM

PR #128, opened 2016-03-16
Use parseInt instead of Number to handle empty string as NaN

### LINKED ISSUE #127: Overlay z-index being set to -1
When a dialog is put into the DOM at the same time as it is opened, the overlay z-index calculation receives "" as the current z-index.
`_getOverlayZ` calculates `var z1 = Number(window.getComputedStyle(overlay).zIndex);` which evaluates to 0 for "" which passes the === check and and hence turns into -1 for the overlay z-index down the line.
This issue can be addressed by changing the calculation to use parseInt() instead of Number(), which would evaluate "" as NaN and let minimumZ be used in such a case.

lkovaleva:
Fixes #127 since Number("") evaluates to 0 whereas parseInt("") evaluates to NaN which should be the expected behavior here.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -243,7 +243,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _getOverlayZ: function(overlay) {
       var z = this._minimumZ;
       if (overlay) {
-        var z1 = Number(window.getComputedStyle(overlay).zIndex);
+        var z1 = parseInt(window.getComputedStyle(overlay).zIndex);
         // Check if is a number
         // Number.isNaN not supported in IE 10+
         if (z1 === z1) {


valdrinkoshi:
Hi @lkovaleva, thanks for the contribution! Looks good, could you also update the tests so we avoid regressions in the future? Thanks!

lkovaleva:
Thanks. I'm not really familiar with Polymer testing. How would I go about setting up a test for this? I don't really know how I could make it have empty string value and I don't see any plain function unit test to just pass "" and check the output. Any help would be appreciated, thanks.

valdrinkoshi:
when you run `polyserve` you can see the documentation & demo at http://localhost:8080/components/iron-overlay-behavior/ and you can run the tests at http://localhost:8080/components/iron-overlay-behavior/test/

To add the test, reproduce the steps you described in the issue in `test/iron-overlay-behavior.html` (check the other tests for reference). I guess you should append a new overlay object on body, call `open()` and after it is opened, check its z-index. Maybe is better to create a new suite called `imperative overlay creation` where the `setup` will be creating and adding the overlay to body, and `teardown` would be removing the overlay from `body`.
Let me know if this makes sense, happy to help further.

valdrinkoshi:
Closing this as it's 2 months old & with pending response.

PR #130, opened 2016-03-17
avoid duplicates when adding overlays

### LINKED ISSUE #129: manager _overlays has duplicates
When `opened` changes quickly without giving time to animations to complete, we end up having duplicates in the `_manager._overlays` array.
http://jsbin.com/qereval/8/edit?html,console,output

valdrinkoshi:
Fixes #129 by ensuring `_overlays` gets updated synchronously when an `overlay.opened` changes.

Added new method to the manager, `bringToFront` which is useful to bring an overlay on top; it accepts both numbers (more efficient) and elements (more ergonomic for the Polymer user) as input.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -78,10 +78,55 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     /**
+     * Brings the overlay at the specified index to the front.
+     * @param {number} i
+     * @private
+     */
+    _bringOverlayAtIndexToFront: function(i) {
+      var overlay = this._overlays[i];
+      var lastI = this._overlays.length - 1;
+      // If already the top element, return.
+      if (!overlay || i === lastI) {
+        return;
+      }
+      // Update z-index to be on top.
+      var minimumZ = Math.max(this.currentOverlayZ(), this._minimumZ);
+      if (this._getZ(overlay) <= minimumZ) {
+        this._applyOverlayZ(overlay, minimumZ);
+      }
+
+      // Shift other overlays behind the new on top.
+      while (i < lastI) {
+        this._overlays[i] = this._overlays[i + 1];
+        i++;
+      }
+      this._overlays[lastI] = overlay;
+    },
+
+    /**
+     * Adds the overlay and updates its z-index if it's opened, or removes it if it's closed.
+     * Also updates the backdrop z-index.
+     * @param {Element} overlay
+     */
+    addOrRemoveOverlay: function(overlay) {
+      if (overlay.opened) {
+        this.addOverlay(overlay);
+      } else {
+        this.removeOverlay(overlay);
+      }
+      this.trackBackdrop();
+    },
+
+    /**
      * Tracks overlays for z-index and focus management.
      * @param {Element} overlay
      */
     addOverlay: function(overlay) {
+      var i = this._overlays.indexOf(overlay);
+      if (i >= 0) {
+        this._bringOverlayAtIndexToFront(i);
+        return;
+      }
       var minimumZ = Math.max(this.currentOverlayZ(), this._minimumZ);
       this._overlays.push(overlay);
       var newZ = this.currentOverlayZ();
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
Tests failing because safari7, I've run them locally with this patch and they're green https://github.com/PolymerElements/test-fixture/pull/32

iron-overlay-manager.html
```diff
@@ -78,10 +78,58 @@
     },
 
     /**
+     * Brings the overlay at the specified index to the front.
+     * @param {number|Element} i
+     */
+    bringToFront: function(i) {
```
cdata: I think it would be cleaner if we had explicit methods for different argument types. e.g., `bringToFront(HTMLElement)` and `brightToFrontByIndex(Number)` or something along those lines. WDYT?

valdrinkoshi: Uhm, what if i make this private & expose it when someone actually asks for it? For what concerns this PR, we need the method `bringToFrontByIndex` (for perf reasons since we already have the index when we use it).
Anyways yes, I also agree we could be more explicit by name, or by accepting an object as input.

iron-overlay-manager.html
```diff
@@ -78,10 +78,58 @@
...
+        i = this._overlays.indexOf(i);
+      }
+      var overlay = this._overlays[i];
+      var lastI = this._overlays.length - 1;
+      // If already the top element, return.
+      if (!overlay || i === lastI) {
+        return;
+      }
+      // Update z-index to be on top.
+      var minimumZ = Math.max(this.currentOverlayZ(), this._minimumZ);
```
cdata: I think `currentOverlayZ` will effectively floor at `this._minimumZ` https://github.com/PolymerElements/iron-overlay-behavior/blob/issue-129/iron-overlay-manager.html#L249-L260

iron-overlay-manager.html
cdata: Overall, I think it's most hygienic not to define APIs that accept dynamic types unless the type is not important to the implementation (e.g., you don't need `typeof` or other reflection features).

I'm happy if you make this private, but also I would prefer if we stick to one type for now unless explicitly needed.

iron-overlay-manager.html
cdata: Ah, nevermind. I thought `_minimumZ` was supposed to update to reflect the highest `zIndex` used.

iron-overlay-manager.html
```diff
@@ -78,10 +78,58 @@
...
+      if (this._getZ(overlay) <= minimumZ) {
+        this._applyOverlayZ(overlay, minimumZ);
+      }
+
+      // Shift other overlays behind the new on top.
+      while (i < lastI) {
+        this._overlays[i] = this._overlays[i + 1];
```
cdata: Why not `splice` and `push`?

```javascript
this._overlays.splice(i, 1); // Removes overlay in place
this._overlays.push(overlay); // Pushes overlay to the end of the array
```

iron-overlay-manager.html
```diff
@@ -78,10 +78,58 @@
...
+      }
+      this._overlays[lastI] = overlay;
+    },
+
+    /**
+     * Adds the overlay and updates its z-index if it's opened, or removes it if it's closed.
+     * Also updates the backdrop z-index.
+     * @param {Element} overlay
+     */
+    trackOverlay: function(overlay) {
```
cdata: This seems like a good method to have, but `track` is an ambiguous verb to me. Given that the behavior is conditional, something like `addOrRemoveOverlay` might be appropriate. WDYT?

iron-overlay-manager.html
```diff
@@ -78,10 +78,58 @@
...
+      }
+      this.trackBackdrop();
+    },
+
+    /**
      * Tracks overlays for z-index and focus management.
      * @param {Element} overlay
      */
     addOverlay: function(overlay) {
+      var i = this._overlays.indexOf(overlay);
```
cdata: Is this `indexOf` necessary? Won't `bringToFront` already perform this test?

iron-overlay-manager.html
valdrinkoshi: :+1: I'll make this private and it will accept only numbers

iron-overlay-manager.html
valdrinkoshi: Mainly for performance on arrays with small size. Inspired from
http://www.redips.net/javascript/array-move/
small dataset test http://jsperf.com/array-prototype-move
big dataset test http://jsperf.com/array-prototype-move/8
WDYT?

iron-overlay-manager.html
valdrinkoshi: Sure, that works. Will update

iron-overlay-manager.html
valdrinkoshi: `bringToFront` accepts a number as input, so this is merely for calling the minimum amount of times `indexOf`  :grin: 
It will be clearer as soon as i update the method name to something like `bringOverlayAtIndexToFront`

valdrinkoshi:
@cdata feedback incorporated, PTAL :)

iron-overlay-manager.html
cdata: It's up to you. FWIW http://jsperf.com/array-prototype-move/24 1 x splice + 1 x push seems to be the same order of magnitude as this verbose alternative, and the cost is so small I doubt anyone will ever notice it in practice. No need to block on this change if you don't want to.

iron-overlay-manager.html
cdata: LGTM :+1:

cdata:
LGTM :+1:

iron-overlay-manager.html
valdrinkoshi: Uhm, our `move2` is 4x faster than `move1` in general conditions (e.g. move an element that has low index), but incredibly faster for elements with high index http://jsperf.com/bring-to-front
This is gonna be a differentiator for features like `always-on-top`, because the top overlay will most likely have a high index in the array.

PR #133, opened 2016-03-18
close backdrop on detached

### LINKED ISSUE #132: when overlay withBackdrop is detached, iron-overlay-backdrop should be synchronously detached too
Backdrop element should be immediately removed from its parent when last overlay is detached:
http://jsbin.com/leboba/1/edit?html,console,output

valdrinkoshi:
Fixes #132 by ensuring `backdropElement` is closed right away on overlay `detached`. Also, `iron-overlay-backdrop` will have `pointer-events: none` when `opened = false`. This allows user interactions while the backdrop is animating :tada:

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -40,12 +40,14 @@ Custom property | Description | Default
       background-color: var(--iron-overlay-backdrop-background-color, #000);
       opacity: 0;
       transition: opacity 0.2s;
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
reviewed offline with @cdata, got his blessing & :+1:

valdrinkoshi:
Tests failing on safari7 due to issues with `<template>`. Verified manually on safari7.

PR #134, opened 2016-03-19
remove a11y-keys behavior from overlay-behavior. Update jsdoc for rootTarget usage

valdrinkoshi:
`iron-overlay-behavior` does not import `iron-a11y-keys-behavior` anymore, neither uses this behavior (key listeners were moved to `iron-overlay-manager`).
Updated jsdoc for how the `rootTarget` is meant to be used in manager's `_onCaptureClick`

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -559,7 +559,7 @@ context. You should place this element as a child of `<body>` whenever possible.
   };
 
   /** @polymerBehavior */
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
@rictic FYI, if you have time to review this.
Tests fail on safari7. Applied locally this patch https://github.com/PolymerElements/test-fixture/pull/32 and tests pass.

rictic:
LGTM

PR #141, opened 2016-03-24
Fixed focus loss on other elements when closing overlay and noAutoFocus is set to true

### LINKED ISSUE #139: focus breaks even with noAutoFocus
I use a paper-dialog to show a paper-input for a search query. 
When there is a paper-toast triggered in the background the input loses focus when the toast hides again.
Wouldn't it make more sense to prevent the blur also if the noAutoFocus attribute is set?
Like this:
```javascript
    _applyFocus: function() {
      if (!this.noAutoFocus) {
        if (this.opened) {
          this._focusNode.focus();
        } else {
          this._focusNode.blur();
          this._focusedChild = null;
          this._manager.focusOverlay();
        }
      }
    },
```

ahedderich:
Fixes #139

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -453,14 +453,14 @@ context. You should place this element as a child of `<body>` whenever possible.
      * @protected
      */
     _applyFocus: function() {
-      if (this.opened) {
-        if (!this.noAutoFocus) {
+      if (!this.noAutoFocus) {
+        if (this.opened) {
           this._focusNode.focus();
+        } else {
+          this._focusNode.blur();
+          this._focusedChild = null;
+          this._manager.focusOverlay();
         }
-      } else {
-        this._focusNode.blur();
-        this._focusedChild = null;
-        this._manager.focusOverlay();
       }
     },
 


ahedderich:
I signed it!

ahedderich:
signed

valdrinkoshi:
Hi @ahedderich, thanks for the contribution! Could you add some unit tests so we avoid any regression?
Thanks!

iron-overlay-behavior.html
```diff
@@ -453,14 +453,14 @@
...
           this._focusNode.focus();
+        } else {
+          this._focusNode.blur();
+          this._focusedChild = null;
+          this._manager.focusOverlay();
         }
-      } else {
-        this._focusNode.blur();
-        this._focusedChild = null;
-        this._manager.focusOverlay();
```
valdrinkoshi: You should still call the `manager.focusOverlay()` if noAutoFocus is true and opened is false, otherwise you break the flow.

valdrinkoshi:
Closing this for now, [see comment](https://github.com/PolymerElements/iron-overlay-behavior/issues/139#issuecomment-201563941)

PR #142, opened 2016-03-25
make _openedChangedAsync a microtask

valdrinkoshi:
The 1ms delay was previously needed for toggling listeners, but since they're now handled in the manager, `_openedChangedAsync` can return to be a micro-task.


cdata:
LGTM :+1:

PR #143, opened 2016-03-25
use path to correctly handle composed tree

valdrinkoshi:
Rather than traversing the parent tree starting from a node, we use `event.path` to have the composed tree, in order to safely recognize who's the deepest overlay in that tree.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -293,15 +293,29 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     /**
+     * Returns the deepest overlay in the path.
+     * @param {?Array<Element>} path
+     * @return {Element|undefined}
+     * @private
+     */
+    _overlayInPath: function(path) {
+      path = path || [];
+      for (var i = 0; i < path.length; i++) {
+        if (path[i]._manager === this) {
+          return path[i];
+        }
+      }
+    },
+
+    /**
      * Ensures the click event is delegated to the right overlay.
      * @param {!Event} event
      * @private
      */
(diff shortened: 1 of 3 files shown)


iron-overlay-manager.html
```diff
@@ -293,15 +293,28 @@
...
+    /**
      * Ensures the click event is delegated to the right overlay.
      * @param {!Event} event
      * @private
      */
     _onCaptureClick: function(event) {
       var overlay = /** @type {?} */ (this.currentOverlay());
-      // Check if clicked outside of any overlay.
-      var target = /** @type {Element} */ (Polymer.dom(event).rootTarget);
-      if (overlay && this._overlayParent(target) !== overlay) {
```
cdata: Should this `_overlayParent` be removed entirely, since the implementation is misleading?

valdrinkoshi: it is still needed in `addOverlay` for:
```javascript
var element = this.deepActiveElement;
overlay.restoreFocusNode = this._overlayParent(element) ? null : element;

valdrinkoshi: Arguably that check could be removed, and we could allow focus to be returned to the previous overlay. This would allow to also remove the call to `this._manager.focusOverlay();` in `iron-overlay-behavior`'s `_applyFocus`.
I'll do this in a separate PR though.

cdata:
LGTM :+1:

valdrinkoshi:
Tests failing on win7 IE10 because of timeout (in saucelabs). Tested locally in win7 IE10 and all is green :green_heart:

PR #144, opened 2016-03-26
Update jsdoc: Accept null & undefined

valdrinkoshi:
Since `event.path` can be either `null` or `undefined`, accept these values as input by setting a default.


rictic:
LGTM

PR #146, opened 2016-03-30
call done

valdrinkoshi:
Fixes tests flakiness


valdrinkoshi:
Reviewed with @notwaldorf offline

PR #148, opened 2016-04-01
listen to touchstart or mousedown

### LINKED ISSUE #147: Open on mousedown causes overlay to cancel (close)
### Description
Overlay cancels (aka closes) if it was opened on a `mousedown` or `mouseup` event.
http://jsbin.com/gajadi/2/edit?html,output
This is because `mousedown` and `mouseup` are triggered before `click` (same for touch events `touchstart` and `touchend` happen before `tap`).

### Expected outcome
Instead of listening for `click` or `tap`, `iron-overlay-manager` should listen for `mousedown` or `touchstart`. Like this, if user starts dragging an area outside the overlay, this would cause the canceling of it.

valdrinkoshi:
Fixes #147 by listening to `touchstart` (touch devices) or `mousedown` rather than `tap/click`.
As a consequence, gestures like dragging outside the dialog cause the dialog to cancel, which is nice for elements like `iron-dropdown` on cases where the user selects text outside the dropdown, causing the dropdown to close.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -38,11 +38,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      */
     this._backdropElement = null;
 
-    // Enable document-wide tap recognizer.
-    Polymer.Gestures.add(document, 'tap', null);
-    // We should be using only 'tap', but this would be a breaking change.
-    var tapEvent = ('ontouchstart' in window) ? 'tap' : 'click';
-    document.addEventListener(tapEvent, this._onCaptureClick.bind(this), true);
+    // Listen to mousedown or touchstart to be sure to be the first to capture
+    // clicks outside the overlay.
+    var clickEvent = ('ontouchstart' in window) ? 'touchstart' : 'mousedown';
+    document.addEventListener(clickEvent, this._onCaptureClick.bind(this), true);
     document.addEventListener('focus', this._onCaptureFocus.bind(this), true);
     document.addEventListener('keydown', this._onCaptureKeyDown.bind(this), true);
   };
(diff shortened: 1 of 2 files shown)


iron-overlay-manager.html
```diff
@@ -38,11 +38,10 @@
...
 
-    // Enable document-wide tap recognizer.
-    Polymer.Gestures.add(document, 'tap', null);
-    // We should be using only 'tap', but this would be a breaking change.
-    var tapEvent = ('ontouchstart' in window) ? 'tap' : 'click';
-    document.addEventListener(tapEvent, this._onCaptureClick.bind(this), true);
+    // Listen to mousedown or touchstart to be sure to be the first to capture
+    // clicks outside the overlay.
+    var clickEvent = ('ontouchstart' in window) ? 'touchstart' : 'mousedown';
+    document.addEventListener(clickEvent, this._onCaptureClick.bind(this), true);
```
cdata: Will this do the right thing on a device that supports touch but is only currently being interacted with via mouse (like the touchpad on a Chromebook Pixel or MS Surface)?

test/iron-overlay-behavior.html
```diff
@@ -242,6 +248,21 @@
           });
         });
 
+        test('open overlay on mousedown does not close it', function(done) {
```
cdata: Ahh, so it's a feature, not a bug? :sake:

iron-overlay-manager.html
valdrinkoshi: At the very least, it will behave as it was before, when we were listening for either `tap` or `click`, not both..
I checked the jsbin of the reported bug on Android & iOs http://jsbin.com/gajadi/2/edit?html,output and it can't be reproduced on the device. I suspect that's because of `Polymer.Gestures.add(document, 'tap', null);` (which I removed here, since strictly speaking `iron-overlays` don't need it anymore). In a generic Polymer app though this gesture will be enabled by other elements that rely on it.

test/iron-overlay-behavior.html
valdrinkoshi: indeed :beers:

cdata:
LGTM :+1:

Issue #149, opened 2016-04-03
The on-screen keyboard incorrectly sizes iron-overlay-backdrop


jdbence:
### Description
The `iron-overlay-backdrop` is incorrectly sized when the on-screen keyboard is shown. It is the size of the viewport but it should be the size of the page.
```js
// what I'm doing now to resolve the issue
iron-overlay-backdrop {
 /*width: 100vw;*/
 /*height: 100vh;*/
 width: 100%;
 height: 100%;
}
```
### Expected outcome
The `iron-overlay-backdrop` should cover the entire page

### Actual outcome
The `iron-overlay-backdrop` is only the size of the viewport

### Steps to reproduce
1. Create a `paper-dialog` with an `input` element
2. Open the page on a iPad browser
3. Tap the input element

### Browsers Affected

- [x] Chrome (iPad)
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

valdrinkoshi:
Hi @jdbence, I tested this http://jsbin.com/vukusa/2/edit?html,output on Chrome 46 and Chrome 49 for iPad and could not reproduce the issue.
Can you provide a jsbin to show the bug?

jdbence:
Sure, I modified your example by adding a class to the body.
[http://jsbin.com/votavi](http://jsbin.com/votavi)

valdrinkoshi:
Thanks!

PR #150, opened 2016-04-03
fix: bug with iron-overlay-backdrop when on-screen keyboard is shown

jdbence:
Any reason why `vh` was used instead of `%`?
Should resolve issue #149

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -22,7 +22,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <script>
       WCT.loadSuites([
         'iron-overlay-behavior.html',
-        'iron-overlay-behavior.html?dom=shadow'
+        'iron-overlay-behavior.html?dom=shadow',
+        'iron-overlay-backdrop.html',
+		'iron-overlay-backdrop.html?dom=shadow',
       ]);
     </script>
 
--- a/test/index.html
+++ b/test/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html><!--
 @license
-Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
 This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
 The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
 The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
@@ -24,10 +24,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         'iron-overlay-behavior.html',
         'iron-overlay-behavior.html?dom=shadow',
         'iron-overlay-backdrop.html',
-		'iron-overlay-backdrop.html?dom=shadow',
+        'iron-overlay-backdrop.html?dom=shadow',
       ]);
     </script>
-
-  
 
 </body></html>
--- a/test/index.html
+++ b/test/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html><!--
 @license
-Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
Thanks for the contribution @jdbence 
Could you add a test so we avoid any future regression? The test should check that the backdrop bounds match the body bounds (make the body scrollable as you did in your example on the issue)

test/iron-overlay-backdrop.html
```diff
@@ -0,0 +1,89 @@
...
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1, user-scalable=yes">
+
+    <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+    <script src="../../web-component-tester/browser.js"></script>
+    <link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
+    <link rel="import" href="test-overlay.html">
+  	<style>
+  		html,
```
valdrinkoshi: Indentation wrong

test/iron-overlay-backdrop.html
```diff
@@ -0,0 +1,89 @@
...
+  	    body {
+  		  margin: 0;
+  		  width: 100%;
+  		  height: 100%;
+  		  min-width: 0;
+  		}
+  		.sizer {
+  			width: 4000px;
```
valdrinkoshi: Indentation wrong

test/iron-overlay-backdrop.html
```diff
@@ -0,0 +1,89 @@
...
+    <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+    <script src="../../web-component-tester/browser.js"></script>
+    <link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
+    <link rel="import" href="test-overlay.html">
+  	<style>
+  		html,
+  	    body {
+  		  margin: 0;
+  		  width: 100%;
+  		  height: 100%;
```
valdrinkoshi: Is this really needed? We want body to be scrollable, so we should not set the height, right?

valdrinkoshi:
Comments on the test, which needs to be formatted with the right indentation across.
Thanks for doing this @jdbence!

valdrinkoshi:
Hi @jdbence, your test is not being run, you need to import it in test/index.html
Or just add that suite and test fixture into test/iron-overlay-behavior.html

jdbence:
Hi @valdrinkoshi,  I've updated the index file to include the test suite.

test/index.html
```diff
@@ -22,7 +22,9 @@
     <script>
       WCT.loadSuites([
         'iron-overlay-behavior.html',
-        'iron-overlay-behavior.html?dom=shadow'
+        'iron-overlay-behavior.html?dom=shadow',
+        'iron-overlay-backdrop.html',
+		'iron-overlay-backdrop.html?dom=shadow',
```
valdrinkoshi: (sorry but) indentation's wrong here :x

test/iron-overlay-backdrop.html
```diff
@@ -0,0 +1,94 @@
+<!doctype html>
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
```
valdrinkoshi: (c) 2016

valdrinkoshi:
@jdbence I left 2 comments..Almost there :)

test/index.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html><!--
 @license
-Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
valdrinkoshi: wrong file, this should remain 2015.

valdrinkoshi:
@jdbence test/iron-overlay-backdrop.html should be @2016 as it was created now. test/index.html should stay @2015

jdbence:
Got it. That makes more sense.

valdrinkoshi:
LGTM :+1: 
Thanks for the contribution @jdbence! :tada:

PR #151, opened 2016-04-04
ensure reflow for smooth animations

valdrinkoshi:
Fixes animation issue for `paper-toast` and `iron-dropdown` happening with the latest `iron-overlay-behavior` (on master). `paper-toast` would not translate when opened, and `iron-dropdown` would blink on safari.

Added `use strict`; did the `display` flipping only on opened (saves a `refit`!). Moved the `_prepareRenderOpened` method (which does the `refit` & `display` flipping) into the `requestAnimationFrame`, for optimal performance.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -15,6 +15,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="iron-overlay-manager.html">
 
 <script>
+// IIFE to help scripts concatenation.
+(function() {
+'use strict';
 
 /**
 Use `Polymer.IronOverlayBehavior` to implement an element that can be hidden or shown, and displays
@@ -321,22 +324,22 @@ context. You should place this element as a child of `<body>` whenever possible.
 
       this.__isAnimating = true;
 
-      if (this.opened) {
-        this._prepareRenderOpened();
-      }
-
       // requestAnimationFrame for non-blocking rendering
       if (this.__openChangedAsync) {
         cancelAnimationFrame(this.__openChangedAsync);
       }
-      this.__openChangedAsync = requestAnimationFrame(function() {
-        this.__openChangedAsync = null;
-        if (this.opened) {
-          this._renderOpened();
-        } else {
-          this._renderClosed();
+      if (this.opened) {
+        if (this.withBackdrop) {
+          this.backdropElement.prepare();
         }
-      }.bind(this));
+        this.__openChangedAsync = requestAnimationFrame(function() {
+          this.__openChangedAsync = null;
+          this._prepareRenderOpened();
+          this._renderOpened();
+        }.bind(this));
+      } else {
+        this._renderClosed();
+      }
     },
 
     _canceledChanged: function() {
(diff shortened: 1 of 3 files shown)


iron-overlay-behavior.html
```diff
@@ -15,7 +15,7 @@
 <link rel="import" href="iron-overlay-manager.html">
 
 <script>
-
+'use strict';
```
cdata: Can you wrap this in a IIFE? For the most part it doesn't matter, but in some cases not doing this will mess up script concatenation.

valdrinkoshi: Which cases would that be needed? So I can document it properly.

cdata: If vulcanization or similar process would cause two scripts to be merged, it could end up like this:

```javascript
// Script 1:
(function() { /* etc ... */ })();
// Script 2:
'use strict';
// etc ...
```

In this case, the `use strict`; for "Script 2" would have no effect. If you wrap this script in an IIFE, then it becomes:

```javascript
// Script 1:
(function() { /* etc ... */ })();
// Script 2:
(function() {
  'use strict';
  // etc ...
})();
```

Then the `'use strict';` will always appear at the top of a function scope, so it will have an effect.

cdata:
This looks fine to me. Two things worry me, though:

1. I wish there was a regression test. I know it's not easy, but it's worth thinking about why this is hard and how we can approach the problem.
2. I would like to confirm that all of the elements that depend on this behavior look right with this change.

WDYT?

iron-overlay-behavior.html
valdrinkoshi: I see. Well, shouldn't this be handled by the vulcanization script? I don't go crazy about adding an IFEE just for concatenation :/

valdrinkoshi:
Since testing is hard, I've ensured this works on the other elements by testing manually (Safari, Chrome, Firefox, IE). Will squeeze my brain a bit more to figure out how to write a regression test ;)

iron-overlay-behavior.html
cdata: This is an old argument, and a bike shed trap. For example, please refer to [this issue](https://github.com/twbs/bootstrap/issues/3057) about semi-colons (or lack thereof) in Bootstrap sources. If you ask me, it is not the responsibility of tools to make guesses about how the code should be formatted, but this is my opinion. Please use your best judgement!

valdrinkoshi: word. Let's add the IFEE :beers:

test/iron-overlay-behavior.html
```diff
@@ -1039,6 +1039,23 @@
 
       });
 
+      suite('animations', function() {
+
+        test('overlay animations correctly triggered', function(done) {
```
valdrinkoshi: @cdata found a cool way to test this :tada:

valdrinkoshi:
@cdata, ok this time tests will pass! Apparently safari 7 doesn't handle very well `transform: translateY(50vh)`, but likes `transform: translateY(300px)` :crystal_ball:

test/iron-overlay-behavior.html
cdata: > cool

I'm guessing that was sarcastic, but thanks for going the distance :beers: I really appreciate it.

valdrinkoshi: eheh sarcastic..yes and no! `document.elementFromPoint` is actually a cool way to see what's where without causing a reflow :grin:

cdata:
You are a testing champion :+1: LGTM!

PR #155, opened 2016-04-11
[WIP] overcome stacking context

valdrinkoshi:
## Update

Polymer 2.0-preview will no longer produce type-extension elements (aka `is="..."`) [(rationale here)](https://github.com/Polymer/polymer/tree/2.0-preview#v1-custom-elements), hence we won't merge this PR as it introduces new custom elements extending `<template>`.

Further work is being done in https://github.com/PolymerLabs/iron-overlay/tree/use-2.0

-----

### The problem

Overlays that are contained into a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) that is not the main one (the one of `document.body`) can have positioning problems, cropping, be non-interactive ([see example](http://jsbin.com/lemufe/4/edit?html,output)). This is a known limitation mentioned also in the [docs](https://elements.polymer-project.org/elements/iron-overlay-behavior):

> You must ensure no element has a stacking context with a higher z-index than its parent stacking context. You should place this element as a child of &lt;body&gt; whenever possible.

These are some stacking-context-related issues from components relying on `iron-overlay-behavior`:
https://github.com/PolymerElements/paper-dialog/issues/7
https://github.com/PolymerElements/polymer-starter-kit/issues/154
https://github.com/PolymerElements/paper-dialog/issues/79
https://github.com/PolymerElements/paper-dialog/issues/44
https://github.com/PolymerElements/paper-drawer-panel/issues/122
### Proposal
- Contain overlays in templates so they are attached only once in the document
- Stamp them and append them into The Right Place™
- Allow content distribution into the overlays contained in an element's shadowDOM (e.g. [`paper-menu-button`](https://github.com/PolymerElements/paper-menu-button/blob/master/paper-menu-button.html) use case)

This is achieved with 4 components:
#### iron-overlay-template

It is a template that takes care of stamping the overlay and delegating the attach to the listeners of the event `iron-overlay-template-attach`. Styles can be added into the template, they’ll be appended into the shadowDOM as well. It provides a handy pointer to the overlay instance (`myTemplate.overlay`).
#### iron-overlay-container

It is the Right Place™. It listens to the `iron-overlay-template-attach` event and appends the overlay in its shadowDOM (so that style encapsulating is guaranteed). You can declare multiple instances (e.g. 1 container for toasts and 1 for dialogs).
#### iron-overlay-content-template

It is a template that stamps and appends its content into `contentParent`. This is needed by elements that host an overlay into their shadowDOM and want to distribute content in them, e.g. `paper-menu-button` needs to distribute dropdown content into the `iron-dropdown` declared in its shadowDOM.
#### iron-overlay-host-behavior

It searches for the `iron-overlay-template` in the shadowDOM and sets it as the `contentParent` of the `iron-overlay-content-template` in the lightDOM (the distributed one).

This changes the markup of the user of your element like this:

``` html
<!-- this div creates a new stacking context -->
<div style="transform: translateZ(0)">
  <!-- it will have stacking context issues -->
  <paper-dialog id="dialog" modal></paper-dialog>
  <button onclick="dialog.open()">open</button> 

  <!-- it will be appended in the iron-overlay-container! -->
  <template id="dialogTemplate" is="overlay-template">
    <paper-dialog modal></paper-dialog>
  </template>
  <button onclick="dialogTemplate.overlay.open()">open</button> 

  <!-- it will have stacking context issues -->
  <paper-menu-button>
    <button class="dropdown-trigger">open</button>
    <div class="dropdown-content">Content</div>
  </paper-menu-button>

  <!-- assuming it implements iron-overlay-host-behavior -->
  <paper-menu-button>
    <button class="dropdown-trigger">open</button>
    <!-- it will be stamped into the overlay -->
    <template is="overlay-content">
      <div class="dropdown-content">Content</div>
    </template>
  </paper-menu-button>
</div>
<!-- this is where I want the overlays to be contained -->
<iron-overlay-container></iron-overlay-container>
```
### Styling

The content of an `iron-overlay-template` will be added into an `iron-overlay-container`'s root, so one could define styles into the template:

``` html
<template is="overlay-template">
  <style> 
    paper-dialog { background-color: yellow; }
  </style>
  <paper-dialog modal></paper-dialog>
</template>
```

Beware that this applies to **all** the `paper-dialog`s that will be contained in an `iron-overlay-container`. 

If you are an element author which hosts an overlay into its shadowDOM, you might want to namespace it, so that its styles won't conflict with other overlays of the same kind:

``` html
<dom-module id="my-element">
<template> 
  <content></content>
  <template is="overlay-template">
    <style>
      paper-dialog.my-element-dialog { background-color: orange; }
    </style>
    <paper-dialog class="my-element-dialog" modal></paper-dialog>
  </template>
</template>
...
```

Note for element authors using `iron-host-behavior`: Polymer currently removes all `style` nodes from a template, so as a temporary solution you'll need to use `<script type="css">` instead of `<style>`.


iron-overlay-container.html
```diff
@@ -0,0 +1,139 @@
...
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="iron-overlay-manager.html">
+
+<!--
+`iron-overlay-container` stamps overlays to `document.body` in order to avoid
```
notwaldorf: 'stamps' is a bit of a pedantic term. Can we use something more user friendly? Adds? Creates?

iron-overlay-container.html
```diff
@@ -0,0 +1,139 @@
...
+any stacking context issue that would cause clipping or rendering behind the backdrop element.
+The rest of the content is distributed within `iron-overlay-container`.
+Mark the template containing the overlay with the attribute `overlay`.
+
+Example:
+
+    <iron-overlay-container>
+      <template overlay>
```
notwaldorf: I think we should use a class here instead of an attribute. See https://github.com/PolymerElements/paper-elements/issues/23

iron-overlay-container.html
```diff
@@ -0,0 +1,139 @@
...
+        <overlay-impl>
```
notwaldorf: It would be nice if the docs had real elements in them; like 2 examples, one using paper-dialog, and one using iron-dropdown. With words like "If you wanted to add a dialog to your page, you could use...", same for a dropdown, and then maybe a "This can also be used with arbitrary elements" (and then show the example you have now)

valdrinkoshi:
@notwaldorf feedback incorporated!

@cdata FYI, here's the PR for overlays 2.0

chisholmd:
Interesting discussion but not much help for a beginner like me.  I am not sure what steps are sufficient to implement this work around.    Any advice?

zixsma:
@valdrinkoshi Any update on this one?

Tigerino:
Any update?

valdrinkoshi:
@Tigerino @zixsma sorry for the late reply. I've updated this PR's description with more info. In short with Polymer 2.0 we'll be able to extend from another custom element but not from the native elements like `template, input, a` etc, hence the idea is to have a new custom element, [`iron-overlay`](https://github.com/PolymerLabs/iron-overlay/tree/use-2.0), which knows how to handle a `<template>` child (aka stamp it in a stacking context safe place).

valdrinkoshi:
Will close this PR but keep the branch 👌

PR #156, opened 2016-04-14
Reference global window explicitly

garlicnation:


valdrinkoshi:
Could you also add `window.` to `requestAnimationFrame`?

valdrinkoshi:
Closing as already handled in #157

PR #157, opened 2016-04-14
Fix closure compiler issues

valdrinkoshi:
Closure compiler doesn't recognize `cancelAnimationFrame`, so change it to `window.cancelAnimationFrame` (while at it, also `window.requestAnimationFrame`).

Also, the compiler won't recognize the property `alwaysOnTop` for `overlay` (which is of type `Element`), so create a separate method `_shouldBeBehindOverlay` that will tell the compiler to not check the type. While at it, updated the `@param` of other methods to mandate a non-null input.


valdrinkoshi:
@garlicnation PTAL

garlicnation:
@valdrinkoshi LGTM, but we'll want to quickly merge a more type-safe solution.

PR #159, opened 2016-04-16
defer _renderOpen/_renderClose after attached

### LINKED ISSUE #158: call requestAnimationFrame before element is attached
the behavior should wait to be attached before performing the open changed.

valdrinkoshi:
Fixes #158.
`iron-overlay-behavior` was calling `requestAnimationFrame` before it was actually attached to the dom. This would put it in a weird status that might cause wrong bounds calculation (e.g. for `iron-dropdown`). Since elements extending this behavior might memoize position information to save relayouts, those memoized values would end up being wrong (because calculated before the element is actually attached). These wrong calculations end up being different according to the browser, so long story short: wait to be attached, then render opened!

### DIFF
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -155,6 +155,21 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           assert.equal(getComputedStyle(overlay).display, 'none', 'overlay starts hidden');
         });
 
+        test('_renderOpened called only after is attached', function(done) {
+          var overlay = document.createElement('test-overlay');
+          // The overlay is ready at this point, but not yet attached.
+          var spy = sinon.spy(overlay, '_renderOpened');
+          // This triggers _openedChanged.
+          overlay.opened = true;
+          // Even if not attached yet, overlay should be the current overlay!
+          assert.equal(overlay, overlay._manager.currentOverlay(), 'currentOverlay ok');
+          // Wait long enough for requestAnimationFrame callback.
+          overlay.async(function() {
+            assert.isFalse(spy.called, '_renderOpened not called');
+            done();
+          }, 100);
+        });
+
         test('overlay open/close events', function(done) {
           var nevents = 0;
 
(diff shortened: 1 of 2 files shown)


test/iron-overlay-behavior.html
```diff
@@ -155,6 +155,21 @@
           assert.equal(getComputedStyle(overlay).display, 'none', 'overlay starts hidden');
         });
 
+        test('_renderOpened called only after is attached', function(done) {
+          var overlay = document.createElement('test-overlay');
+          // overlay is ready at this point, but not yet attached.
+          var spy = sinon.spy(overlay, '_renderOpened');
+          // this triggers _renderOpened
```
notwaldorf: nit: Capitalize, here and below, since you do that in the middle (and probably everywhere else in the file)

test/iron-overlay-behavior.html
```diff
@@ -155,6 +155,21 @@
...
+          overlay.opened = true;
+          // Even if not attached yet, overlay should be the current overlay!
+          assert.equal(overlay, overlay._manager.currentOverlay(), 'currentOverlay ok');
+          // wait for requestAnimationFrame to kick.
+          overlay.async(function() {
```
notwaldorf: Hmm, this is a weird test. You're basically saying "this won't be called in 100ms", but what if it's called in 101ms? :(

valdrinkoshi: Uhm, I could probably add the overlay into body & check that `_renderOpened` was called only once..! Will do it!

valdrinkoshi: Here I'm waiting "long enough" to see what that `opened` change has triggered. I don't see another way to check for this, either increase the time or use `requestAnimationFrame`..I prefer the timeout of 100ms to the latter.
PS: when I run this test without my fix I get it failing.

test/iron-overlay-behavior.html
valdrinkoshi: done!

notwaldorf:
LGTM. That test is still weird to me, but I understand how there's nothing else really we can do :)

PR #161, opened 2016-04-19
withBackdrop prevents click outside

### LINKED ISSUE #160: An anchor link behind a backdrop reacts with tap on closing the backdrop
### Description
An anchor link behind a backdrop unintentionally reacts with a tap action as soon as the backdrop has closed.

I guess this is a regression related to #148.

### Expected outcome
The anchor does not react with the tap while the backdrop is closing.

### Actual outcome
The anchor reacts with the tap, and jumps to the link.

### Live Demo
http://jsbin.com/hebacicize/edit?html,console,output

### Steps to reproduce

1. Tap a `paper-button` in the page. (will open a backdrop)
2. Tap the backdrop just around on a link.

### Browsers Affected
- [x] Moboile Chrome
- [x] Moboile Safari

valdrinkoshi:
Fixes #160 by calling `event.preventDefault()` for overlays with backdrop. This ensures that click/tap events don't go beyond the `backdropElement`.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -339,6 +339,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var overlay = /** @type {?} */ (this.currentOverlay());
       // Check if clicked outside of top overlay.
       if (overlay && this._overlayInPath(Polymer.dom(event).path) !== overlay) {
+        if (overlay.withBackdrop) {
+          // There's no need to stop the propagation as the backdrop element
+          // already got this mousedown/touchstart event. Calling preventDefault
+          // on this event ensures that click/tap won't be triggered at all.
+          event.preventDefault();
+        }
         overlay._onCaptureClick(event);
       }
     },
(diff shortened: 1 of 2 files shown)


iron-overlay-behavior.html
```diff
@@ -492,6 +492,10 @@
      */
     _onCaptureClick: function(event) {
       if (!this.noCancelOnOutsideClick) {
+        if (this.withBackdrop) {
+          // Avoid click to go beyond the backdrop element.
+          event.preventDefault();
```
notwaldorf: Do you want that or `stopImmediatePropagation`?

valdrinkoshi: Since the backdrop will already get this `mousedown/touchstart` event while still opaque/clickable, there's no need to stop the propagation.
`preventDefault` ensures that the following events `click/tap` won't happen at all 😁

valdrinkoshi: (i should probably update the comment then!)

valdrinkoshi: question: do you think this is a duty of the manager? it is the one listening for `touchstart` after all..WDYT @notwaldorf ?

valdrinkoshi:
@notwaldorf moved the check into the manager & updated the comment. PTAL

notwaldorf:
LGTM

PR #162, opened 2016-04-21
Move backdrop updates to manager

valdrinkoshi:
`iron-overlay-backdrop` now allows to set `opened` synchronously, and animation to `opened` state is handled with a css class instead of attribute, using `requestAnimationFrame` for better performance.
Removed the dependency to `iron-overlay-manager` as it was needed to limit when the backdrop can be opened or not. This part is controlled by the manager itself now.

`iron-overlay-behavior` delegates the updates of `backdropElement`'s `opened` state to `iron-overlay-manager`.

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -30,7 +30,6 @@ Custom property | Description | Default
 <dom-module id="iron-overlay-backdrop">
 
   <style>
-
     :host {
       position: fixed;
       top: 0;
@@ -41,17 +40,14 @@ Custom property | Description | Default
       opacity: 0;
       transition: opacity 0.2s;
       pointer-events: none;
-
       @apply(--iron-overlay-backdrop);
     }
 
-    :host([opened]) {
+    :host(.opened) {
       opacity: var(--iron-overlay-backdrop-opacity, 0.6);
       pointer-events: auto;
-
       @apply(--iron-overlay-backdrop-opened);
     }
-
   </style>
 
   <template>
@@ -61,9 +57,6 @@ Custom property | Description | Default
 </dom-module>
 
 <script>
-
-(function() {
-
   Polymer({
 
     is: 'iron-overlay-backdrop',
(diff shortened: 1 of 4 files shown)


iron-overlay-backdrop.html
```diff
@@ -74,10 +67,10 @@
        * Returns true if the backdrop is opened.
        */
       opened: {
-        readOnly: true,
-        reflectToAttribute: true,
```
cdata: Is this not a breaking change?

cdata:
This mostly looks good, although I'm concerned about broken API. I know using a class is better for performance, but is it critical for this change?

valdrinkoshi:
@cdata feedback incorporated, not a breaking change anymore 🍻 PTAL

iron-overlay-backdrop.html
```diff
@@ -61,8 +57,9 @@
 </dom-module>
 
 <script>
-
+// IIFE to help scripts concatenation.
```
cdata: This comment isn't necessary.

iron-overlay-backdrop.html
```diff
@@ -116,33 +128,53 @@
...
-        }
+        this.setOpened(false);
+      }
+    },
+
+
+    /**
+     * Updates opened.
+     */
+    setOpened: function(opened) {
```
cdata: At this point, `opened` technically isn't `readOnly` anymore :P

iron-overlay-backdrop.html
```diff
@@ -87,8 +85,22 @@
 
     },
 
+    get _noAnimation() {
+      var cs = getComputedStyle(this);
+      return (cs.transitionDuration === '0s' || cs.opacity == 0);
```
cdata: Are these the only conditions?

iron-overlay-backdrop.html
valdrinkoshi: True, but it will make the implementation of `_openedChanged` uglier :s

iron-overlay-backdrop.html
cdata: Also, does `opacity === 0` really imply no animation?

iron-overlay-backdrop.html
valdrinkoshi: 👍 will update

valdrinkoshi:
@cdata feedback incorporated, PTAL

cdata:
LGTM 👍

PR #165, opened 2016-04-28
fix: make overlay display over backdrop, by appending overlay to body

timblack1:
Fixes https://github.com/PolymerElements/paper-dialog/issues/7
Fixes https://github.com/PolymerElements/polymer-starter-kit/issues/154
Fixes (maybe) https://github.com/PolymerElements/paper-dialog/issues/79
Fixes (maybe) https://github.com/PolymerElements/paper-dialog/issues/44
Fixes (maybe) https://github.com/PolymerElements/paper-drawer-panel/issues/122
Related to https://github.com/PolymerElements/iron-overlay-behavior/pull/72
Related to https://github.com/PolymerElements/iron-overlay-behavior/pull/86
Related to https://github.com/PolymerElements/iron-overlay-behavior/compare/backdrop-refit
Fixes (maybe) https://github.com/PolymerElements/paper-dialog-behavior/issues/78
Fixes https://github.com/PolymerLabs/app-layout/issues/155

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -88,6 +88,22 @@ context. You should place this element as a child of `<body>` whenever possible.
       },
 
       /**
+       * Set to true for this overlay to remain a child of its original
+       * parent element.
+       *
+       * By default, this overlay element appends both its overlay
+       * and its backdrop as an immediate child of the `body` element.
+       *
+       * If you want to keep the overlay as a child of its original
+       * parent element, set `keepWithParent` to "true".
+       */
+      keepWithParent: {
+        type: Boolean,
+        default: false,
+        notify: true
+      },
+
+      /**
        * Set to true to disable auto-focusing the overlay or child nodes with
        * the `autofocus` attribute` when the overlay is opened.
        */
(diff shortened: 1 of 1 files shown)


timblack1:
I signed it!

valdrinkoshi:
Hi @timblack1, the current proposal has several drawbacks, e.g. it breaks style encapsulation. Also, the fact that you reparent it will cause `detach` and `attach` to be triggered, potentially entering in an infinite loop.

timblack1:
Does https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-backdrop.html#L99 break style encapsulation too?  That is the line which necessitates a fix like the one proposed here.

valdrinkoshi:
Not really, as it is done if you explicitly call the method and if the element doesn't have a parent.
In essence it is suboptimal to attach, detach & attach elsewhere. This would cause query selectors to stop working, it would be a very breaking change.

I'm working on overcoming these stacking context related issues in the stacking-context branch. Happy to discuss the approach in a separate issue!

timblack1:
I don't see an infinite loop in my manual testing of this proposed change.  I look forward to seeing a good resolution of the problem.

valdrinkoshi:
The PR with the proposal is this one https://github.com/PolymerElements/iron-overlay-behavior/pull/155
Closing this PR.

AliMD:
its works but its not a good way for solving problem, its like a hack dom. and maybe make problem with data binding, etc ...

PR #169, opened 2016-05-05
call renderClosed inside requestAnimationFrame too

valdrinkoshi:
Like for `renderOpened`, we want also `renderClosed` to have non-blocking animations, so I moved it into the `requestAnimationFrame`


cdata:
LGTM 👍

PR #171, opened 2016-05-10
force focus wrapping

### LINKED ISSUE #170: Focus isn't wrapping when tabbing (only works with shift-tab)
### Description
Tabbing through a modal paper-dialog does not trap focus. However, if I shift-tab it does.

### Expected outcome

Focus should wrap, if i'm tabbing forward

### Actual outcome

Focus escapes the modal when tabbing forward

### Live Demo
http://jsbin.com/ciqugegulu/1/edit?html,output

Note that autofocus also seems messed up in jsbin but is working for me locally

### Steps to reproduce

Click the open modal button
Click inside the modal to move focus navigation start point
Press tab 3 times and note it does not go back to `decline` button.

### Browsers Affected

- [ x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

valdrinkoshi:
Fixes #170 by preventing event default and forcing the focus to wrap instead of relying on focus change.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -516,6 +516,16 @@ context. You should place this element as a child of `<body>` whenever possible.
       var nodeToCheck = shift ? this.__firstFocusableNode : this.__lastFocusableNode;
       var nodeToSet = shift ? this.__lastFocusableNode : this.__firstFocusableNode;
       if (this.withBackdrop && this._focusedChild === nodeToCheck) {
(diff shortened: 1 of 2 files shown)


notwaldorf:
I guess this is ok. Please write a comment though, since this is kind of weird behaviour, and `event.preventDefault` looks alarming :(

PR #177, opened 2016-05-12
_focusableNodes can be overridden.

### LINKED ISSUE #176: _focusableNodes not called if overridden
### Description
The `IronOverlayBehaviorImpl._focusableNodes` gets called on `attached` instead of the overridden one.
The update of the first & last focusable should probably be deferred only when the dialog is opened, or there should be 2 properties that can be overridden for those.

### Live Demo
http://jsbin.com/xubasi/1/edit?html,output

valdrinkoshi:
Fixes #176 by invoking `_focusableNodes` after the overlay is opened instead of during nodes changes.
What was happening previously is that `_focusedNodes` would be called **before** the getter could be overridden by `_onNodesChange`. 
Now, we update the first & last focusable each time an overlay is opened.
Ideally we should have two properties for first and last focusable nodes and deprecate/remove `_focusableNodes`, but that might be a breaking change, while this can qualify as a patch.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -406,6 +406,12 @@ context. You should place this element as a child of `<body>` whenever possible.
 
       this.notifyResize();
       this.__isAnimating = false;
+
+      // Store it so we don't query too much.
+      var focusableNodes = this._focusableNodes;
+      this.__firstFocusableNode = focusableNodes[0];
+      this.__lastFocusableNode = focusableNodes[focusableNodes.length - 1];
+
       this.fire('iron-overlay-opened');
     },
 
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
@blasten FYI

iron-overlay-behavior.html
```diff
@@ -406,6 +406,12 @@
 
       this.notifyResize();
       this.__isAnimating = false;
+
+      // Store it so we don't query too much.
+      var focusableNodes = this._focusableNodes;
+      this.__firstFocusableNode = focusableNodes[0];
```
blasten: 💥

iron-overlay-behavior.html
```diff
@@ -578,10 +584,6 @@
       if (this.opened && !this.__isAnimating) {
         this.notifyResize();
       }
-      // Store it so we don't query too much.
-      var focusableNodes = this._focusableNodes;
-      this.__firstFocusableNode = focusableNodes[0];
-      this.__lastFocusableNode = focusableNodes[focusableNodes.length - 1];
```
blasten: 👍

cdata:
LGTM 👍

PR #178, opened 2016-05-13
iron-focus-wrap-behavior!

valdrinkoshi:
This behavior allows to wrap the focus within an element without stealing it from anyone, but wrapping it as soon as it gets it.
If the user clicks outside the element, it is not the responsibility of this behavior to block that.


valdrinkoshi:
@frankiefu @blasten @cdata PTAL

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,236 @@
...
+      var node = searchInShadowDOM ? this.root : this;
+      var focusables = Polymer.dom(node).querySelectorAll(selector);
+      if (this.tabIndex >= 0) {
+        // Insert at the beginning because we might have all elements with tabIndex = 0,
+        // and the overlay should be the first of the list.
+        focusables.splice(0, 0, this);
+      }
+      // Sort by tabindex.
+      return focusables.sort(function (a, b) {
+        if (a.tabIndex === b.tabIndex) {
```
blasten: isn't this equivalent to `a.tabIndex - b.tabIndex`?

valdrinkoshi: true!

blasten: Since now the behavior doesn't have a focus listener, if a user were to set `tabindex` on the focusable elements, it'd be technically possible that focus could go outside. Wdyt if the behavior doesn't support that use case?

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,236 @@
...
+        // the document (e.g. to the browser search bar). Similarly, when the
+        // element contains the first focusable element of the document and it's
+        // already focused, pressing Shift+TAB would move the focus outside the
+        // document (e.g. to the browser search bar).
+        // In both cases, we would not receive a focus event, but only a blur.
+        // In order to achieve focus wrapping, we prevent this TAB event and
+        // force the focus. This will also prevent the focus to temporarily move
+        // outside the element, which might cause scrolling.
+        event.preventDefault();
+        nodeToSet.focus();
```
blasten: 👍

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * Searches and sets the `firstFocusable` and `lastFocusable`. This is very
+     * expensive and you're encouraged to set first and last focusable yourself
+     * if you know the content.
+     * @param {boolean=} searchInShadowDOM If true, the search is performed in
+     * the shadow dom.
+     * @protected
+     */
+    _searchAndSetFocusables: function(searchInShadowDOM) {
```
blasten: wdyt about having a `root` instead of `searchInShadowDOM`?  so If I'd optimize the querySelectorAll to the subtree that has the focusable nodes.  I could also do this:

```js
// light DOM
_searchAndSetFocusables(Polymer.dom(this))  
// Shadow DOM
_searchAndSetFocusables(Polymer.dom(this.root))  
```

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,236 @@
...
+        type: Element,
+        value: function() {
+          return this;
+        }
+      },
+      /**
+       * If true, it will wrap the focus on `TAB` or `Shift+TAB` when the focus
+       * is already in the element.
+       */
+      wrapFocus: Boolean
```
blasten: wdyt about:
```js
 wrapFocus = true
 this.addOwnKeyBinding('tab', '__onTab');

 wrapFocus = false
 this.removeOwnKeyBinding('tab', '__onTab');
```

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,236 @@
...
+     * Updates the `storedDeepActiveElement` to the currently focused element.
+     */
+    saveDeepActiveElement: function() {
+      this.__deepActiveElement = this.__getDeepActiveElement();
+    },
+
+    /**
+     * Restores the focus to the `storedDeepActiveElement`.
+     */
+    restoreDeepActiveElement: function() {
```
blasten: 💯

iron-focus-wrap-behavior.html
valdrinkoshi: kudos to that, great idea 💯

iron-focus-wrap-behavior.html
valdrinkoshi: That would require an observer on the property, and we wouldn't benefit too much from removing the key binding (as the event listener will still be active)

iron-focus-wrap-behavior.html
valdrinkoshi: That's true. That could be handled by adding this check in `__onTab`:
```javascript
if (Polymer.dom(event).path.indexOf(this) === -1) { 
  shouldWrap = true;
}
```
This would really be the trapping part..!

valdrinkoshi: Actually, that would not help, since the keyboard event listener is on the element itself, and we don't know where the focus will go after the TAB is done...I'd defer the support of that use case to be handled outside this behavior, WDYT?

iron-focus-wrap-behavior.html
```diff
@@ -0,0 +1,234 @@
...
+      var selector = FOCUSABLE_WITH_DISABLED.join(':not([tabindex="-1"]),') +
+        ':not([tabindex="-1"]),' +
+        FOCUSABLE_WITHOUT_DISABLED.join(':not([disabled]):not([tabindex="-1"]),') +
+        ':not([disabled]):not([tabindex="-1"])';
+
+      var focusables = Polymer.dom(root).querySelectorAll(selector);
+      if (root.hasAttribute('tabindex') && root.tabIndex >= 0) {
+        // Insert at the beginning because we might have all elements with
+        // tabIndex = 0, and the root should be the first of the list.
+        focusables.splice(0, 0, root);
```
blasten: nit: `focusable.unshift(root)`

blasten:
@valdrinkoshi I love this behavior. great work!

MartinMoizard:
Hi @valdrinkoshi, any plan yet to merge this PR? Thanks :)

tgsergeant:
Echoing the comment from above -- @valdrinkoshi, has there been any progress made on this recently? This is looking to be useful for us in Chrome.

valdrinkoshi:
@MartinMoizard @tgsergeant sorry for the late reply!
One of the things this PR solves is to overcome a limitation of `iron-overlay-behavior` to find what are the focusable elements in order to wrap the focus, by exposing 2 properties to the element user (which basically transfers the responsibility but doesn't solve the problem).
I've had time to work on actually solving this problem here https://github.com/PolymerElements/iron-overlay-behavior/pull/200, and I believe that if we can guarantee a proper wrapping of the focus, we don't need to expose new properties.

About having the `focus-wrap-behavior`, I still think there is value in this, but if the `iron-focusables-helper` lands, it can potentially get much simpler, and wouldn't need to expose the `first/lastFocusable` properties; we don't want to ship it now and introduce a breaking change later ;)

Issue #166, opened 2016-05-01
Styling backdrop is impossible/unsatisfyingly difficult


AndreasGalster:
When wanting to style a paper-dialog it's almost impossible to style the backdrop. Since the backdrop get's added to the body, the only way to style it is via :root. This in itself already makes it very difficult to style it differently per element.

Element 1 with paper-dialog might have a black background, element 2 might use a blue background.

Now the only way to somehow conditionally fix that would be by assigning a class perhaps, which doesn't seem to work though. If you do :root.class {} this doesn't work. Using iron-overlay-backdrop {} works again, whereas something like iron-overlay-backdrop.blue {} doesn't work. Now if by some chance you can find a way to make the backdrop to pick up the styling up via the class there's still another issue: You could only add a class to the backdrop once it's in the body. So doing paper-dialog.toggle() and then paper-dialog.backdropElement.classList.add('blue') doesn't even work either because the timing between toggle and add is off. Ok, guess we could prepare the backdrop first and then add the class. (I have not tried this yet but I assume that would work).

I hope you see where this is going. All I wanted to do was color the backdrop and change the opacity for one paper-dialog. I've spent a few hours coming up with a solution (and still don't have one). I'd suggest an improvement here. I have no idea how to improve it though :P.

For anyone wondering, this is a temporary but definitely ugly hack to get a decent result. In the main document:

`
  <style>
    iron-overlay-backdrop.blue[opened] {
      background: #2196f3;
      opacity: 1;
      transition: opacity 1s ease;
      transition: background 0.3s ease;
    }
  </style>
`

valdrinkoshi:
Hi @AndreasGalster, the `iron-overlay-backdrop` is by design a singleton, hence is shared by all the overlays (in fact it gets generated by the `iron-overlay-manager` and appended to `document.body`).
I'd suggest to toggle a css class on the backdrop right before you open the overlay, e.g. http://jsbin.com/hecehib/5/edit?html,output
I noticed that custom properties don't work, @cdata @notwaldorf any ideas why?

valdrinkoshi:
@AndreasGalster, there are 2 ways to tackle this problem:
[...]

PR #180, opened 2016-05-13
backdrop moved in overlay root

valdrinkoshi:
Move the backdrop into the `overlay`'s root. The backdrop now is `position: absolute` and covers the whole container size, and uses `outline` to cover the rest of the screen. To block the interactions with the rest of the screen, it uses a transparent `div`. The interaction with the overlay content is still possible, as the backdrop has `z-index: -1`. 
### Cons

In `dom=shadow` this will be a breaking change, as to style the backdrop it will be necessary to have a selector on the overlay:

``` css
/* new way to style */
simple-overlay {
  --iron-overlay-backdrop-background-color: red;
}
/* old way to style */
body {
  --iron-overlay-backdrop-background-color: red;
}
```
### Pros

This provides more flexibility on styling the backdrop according to the overlay that will be opened: http://jsbin.com/sanupa/1/edit?html,output (this addresses #166)
### Notes

Adding the backdrop into an overlay's root is "better" than adding it to `document.body`, or into any other root, because it stays encapsulated into the overlay scope.

This doesn't really solve the stacking context problem: it will allow to click into the overlay content, but also to interact with the content outside the overlay (see this example http://jsbin.com/gixuki/4/edit?html,output)

### DIFF
--- a/iron-overlay-backdrop.html
+++ b/iron-overlay-backdrop.html
@@ -22,7 +22,7 @@ Custom property | Description | Default
 -------------------------------------------|------------------------|---------
 `--iron-overlay-backdrop-background-color` | Backdrop background color                                     | #000
 `--iron-overlay-backdrop-opacity`          | Backdrop opacity                                              | 0.6
-`--iron-overlay-backdrop`                  | Mixin applied to `iron-overlay-backdrop`.                      | {}
+`--iron-overlay-backdrop`                  | Mixin applied to `iron-overlay-backdrop`.                     | {}
 `--iron-overlay-backdrop-opened`           | Mixin applied to `iron-overlay-backdrop` when it is displayed | {}
 -->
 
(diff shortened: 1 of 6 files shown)


valdrinkoshi:
@notwaldorf @cdata PTAL

notwaldorf:
This says `breaking-change` on it...should I still review it? :)

notwaldorf:
Should we close this and open a 2.0 tagged issue instead? (maybe referencing this PR)

valdrinkoshi:
Moving the backdrop into the overlay would trap the backdrop into the stacking context where the overlay is trapped as well, meaning: if the overlay (e.g. a modal dialog) is cropped by a parent's stacking context (e.g. a containing sliding app drawer), the backdrop will be cropped as well.

Closing this as it doesn't solve the issue.

PR #181, opened 2016-05-13
Revert of #148

### LINKED ISSUE #174: `iron-overlay` is not closed on outside click when device supports touch + mouse. e.g. Windows 10 laptop with touch support.
### Description
`iron-overlay` is not closed on outside click when device supports touch + mouse. e.g. Windows 10 laptop with touch support.

### Expected outcome
`iron-overlay` should be closed on outside click when device supports touch + mouse. For outside touch it's working.

### Actual outcome
`iron-overlay` is not closed on outside click when device supports touch + mouse. For outside touch it's working.

### Steps to reproduce
Sorry, I am skipping this as I guess, title it self is enough explaining issue. 

### Browsers Affected
All

valdrinkoshi:
Revert #148 since it caused more issues (e.g. #174) than solving problems. Had to do this manually as #148 cannot be automatically reverted.

The whole `mousedown/touchstart` change started after I saw that some components would use `mousedown` to trigger the opening of an overlay. Instead, the manager should just listen to `tap` events, and overlays should remove themselves from the manager synchronously when they get closed, and add them asynchronously when they get opened.

If a component really has to open an overlay on `mousedown` (bad idea), it should take care the overlay doesn't get canceled by listening to `iron-overlay-canceled`, something like this http://jsbin.com/juziyow/3/edit?html,output

This also fixes #174.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -316,10 +316,15 @@ context. You should place this element as a child of `<body>` whenever possible.
         return;
       }
 
-      this._manager.addOrRemoveOverlay(this);
-
       if (this.__openChangedAsync) {
         window.cancelAnimationFrame(this.__openChangedAsync);
+      }
+
+      // Synchronously remove the overlay.
+      // The adding is done asynchronously to go out of the scope of the event
+      // which might have generated the opening.
+      if (!this.opened) {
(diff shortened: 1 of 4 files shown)


frankiefu:
LGTM

PR #116, opened 2016-02-24
Click, focus, keydown listeners on manager

valdrinkoshi:
Fixes #109, fixes #117 by handling click and focus listeners on `iron-overlay-manager`.

Previously, each overlay would have these listeners, and these were toggled a tick after `_onOpenedChanged` (to avoid the triggering click event to close them). This was cumbersome because we'd have a lot of listeners, and each overlay would check if it is the `currentOverlay`.

Now, `iron-overlay-manager` will always listen for `click` and `focus`, and delegates the actions only to the current/top overlay (does also some other safety checks). To avoid the immediate closing, we delay the setting of `currentOverlay` for a tick. This fixes also #117.

Finally, we synchronously update the `_overlays`, so that we are sure the array will always contain opened overlays. This allowed to remove `_backdrops`.

`keydown` listener moved to the Manager as well, in order to use `useCapture` and have less listeners.

Some examples:
- [2 nested overlays](http://jsbin.com/xuguxa/1/edit?html,output) - [with the fix](http://jsbin.com/savivac/1/edit?html,output)
- [2 overlays](http://jsbin.com/zipugig/2/edit?html,output) - [with the fix](http://jsbin.com/vebinu/12/edit?html,output)
- [2 dropdowns](http://jsbin.com/reqayid/3/edit?html,output) - [with the fix](http://jsbin.com/dafaho/3/edit?html,output)

PR #183, opened 2016-05-17
Async remove overlay

### LINKED ISSUE #182: close on click closes more than one overlay
### Description
The click event causes more than one overlay to close. That's because `click` is triggered before the "synthetic" `tap` from `Polymer.Gestures`. The adding of the overlay to the manager is done asynchronously, it should be the same for the removal, so that during the `event` scope the current overlay won't change. 

### Expected outcome

Only the top overlay should be closed.

### Actual outcome

2 overlays are closed.

### Live Demo
http://jsbin.com/nesixah/1/edit?html,output

valdrinkoshi:
Fixes #182.

This PR decouples the adding/removing of an overlay into the manager from the focus management (now delegated to the overlay).
It makes the addition/removal of the overlay into the manager asynchronous in order to skip the event that might have caused the opened change.

Since the initial implementation, the adding of the overlay to the manager was synchronous, while its removal asynchronous. In addition, each overlay would [toggle asynchronously listeners](https://github.com/PolymerElements/iron-overlay-behavior/blob/v1.0.0/iron-overlay-behavior.html#L256) for `tap` and `keydown` events. These listeners were asynchronous to **avoid the execution of the callback during the same scope of the event that triggered the open/close** (e.g. `<button onclick="overlay.open()">`). Also, the callbacks would have to check [if the element is the current overlay, or stop the event propagation](https://github.com/PolymerElements/iron-overlay-behavior/blob/v1.0.0/iron-overlay-behavior.html#L379).

This was sub-optimal when having multiple overlays opened, and lead to issues like all overlays would get closed.
The root cause of the problem was the synchronous addition of the overlay, while the listeners where toggled asynchronously. These 2 operations should instead be done together.

To solve these issues, I moved all these listeners to `IronOverlayManager`, which has visibility over who's the current overlay, and delegates the handling of the event to just one overlay (done in #116).
In that PR, the choice was to make addition and removal of the overlay **synchronous**, in order to have the manager's `currentOverlay` synchronously updated.

Instead, this operation should be **async**, as we need to avoid the callback execution during the scope of the triggering event.

That's what is being (finally) done by this PR 🎉

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -145,14 +145,6 @@ context. You should place this element as a child of `<body>` whenever possible.
       },
 
       /**
-       * Is true during the transition from close to open and open to close.
-       */
-      transitioning: {
-        type: Boolean,
-        readOnly: true
-      },
-
-      /**
        * Shortcut to access to the overlay manager.
        * @private
        * @type {Polymer.IronOverlayManagerClass}
@@ -245,6 +237,8 @@ context. You should place this element as a child of `<body>` whenever possible.
     },
 
     ready: function() {
+      // Used to skip calls to notifyResize and refit while the overlay is animating.
+      this.__isAnimating = false;
       // with-backdrop needs tabindex to be set in order to trap the focus.
       // If it is not set, IronOverlayBehavior will set it, and remove it if with-backdrop = false.
       this.__shouldRemoveTabIndex = false;
@@ -340,18 +334,25 @@ context. You should place this element as a child of `<body>` whenever possible.
         return;
       }
 
-      this._setTransitioning(true);
+      this.__isAnimating = true;
 
       // Use requestAnimationFrame for non-blocking rendering.
       this.__openedChangedRAF && window.cancelAnimationFrame(this.__openedChangedRAF);
       this.__openedChangedRAF = window.requestAnimationFrame(function() {
         this.__openedChangedRAF = null;
         if (opened) {
+          // Make overlay visible, then add it to the manager.
           this._prepareRenderOpened();
+          this._manager.addOverlay(this);
+          // Move the focus to the child node with [autofocus].
+          this._applyFocus();
+
           this._renderOpened();
         } else {
-          // Move the focus before actually closing.
+          // Remove overlay, then restore the focus before actually closing.
+          this._manager.removeOverlay(this);
           this._applyFocus();
+
           this._renderClosed();
         }
       }.bind(this));
@@ -390,12 +391,9 @@ context. You should place this element as a child of `<body>` whenever possible.
       this.refit();
       this._finishPositioning();
 
-      // Move the focus to the child node with [autofocus].
-      this._applyFocus();
-
       // Safari will apply the focus to the autofocus element when displayed
       // for the first time, so we make sure to return the focus where it was.
-      if (this.noAutoFocus) {
+      if (this.noAutoFocus && document.activeElement === this._focusNode) {
         this._focusNode.blur();
         this.__restoreFocusNode.focus();
       }
@@ -422,12 +420,8 @@ context. You should place this element as a child of `<body>` whenever possible.
      * @protected
      */
     _finishRenderOpened: function() {
-      this._manager.addOverlay(this);
-      // Focus the child node with [autofocus]
-      this._applyFocus();
-
       this.notifyResize();
-      this._setTransitioning(false);
+      this.__isAnimating = false;
 
       // Store it so we don't query too much.
       var focusableNodes = this._focusableNodes;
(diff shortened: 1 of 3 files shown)


iron-overlay-behavior.html
```diff
@@ -145,6 +145,14 @@
       },
 
       /**
+       * Is true during the transition from close to open and open to close.
+       */
+      transitioning: {
```
notwaldorf: does this need to be a public property?

iron-overlay-behavior.html
```diff
@@ -392,10 +393,11 @@
...
       this._applyFocus();
 
-      // Safari will apply the focus to the autofocus element when displayed for the first time,
-      // so we blur it. Later, _applyFocus will set the focus if necessary.
-      if (this.noAutoFocus && document.activeElement === this._focusNode) {
+      // Safari will apply the focus to the autofocus element when displayed
+      // for the first time, so we make sure to return the focus where it was.
+      if (this.noAutoFocus) {
         this._focusNode.blur();
+        this.__restoreFocusNode.focus();
```
notwaldorf: I remember merging other PRs related to this. Is this still needed?

iron-overlay-behavior.html
```diff
@@ -420,9 +422,12 @@
      * @protected
      */
     _finishRenderOpened: function() {
+      this._manager.addOverlay(this);
+      // Focus the child node with [autofocus]
+      this._applyFocus();
```
notwaldorf: Same question about other PRs as above

iron-overlay-behavior.html
valdrinkoshi: `IronOverlayManager` checks for the `transitioning` property [since ever](https://github.com/PolymerElements/iron-overlay-behavior/commit/09e757a14cd6cab9ba13d961ef38d12f36b57359#diff-3a555b03d77a3dc28aad3488eed08943R73) in order to apply or not the focus, but I can probably remove this and keep the `__isAnimating` property as before.

iron-overlay-behavior.html
valdrinkoshi: This is because of what Safari does when an `autofocus` element is displayed for the first time (see comment at 396). Not only we want to blur the `_focusNode`, but also move the focus to where it was before

valdrinkoshi: I can probably move this logic inside `_applyFocus`..!

valdrinkoshi:
@notwaldorf feedback incorporated, PTAL :)

iron-overlay-behavior.html
```diff
@@ -245,24 +245,28 @@
...
+        this._openedChanged(this.opened);
       }
       this._observer = Polymer.dom(this).observeNodes(this._onNodesChange);
     },
 
     detached: function() {
       Polymer.dom(this).unobserveNodes(this._observer);
       this._observer = null;
-      this.opened = false;
+      this.__openedChangedRAF && window.cancelAnimationFrame(this.__openedChangedRAF);
```
notwaldorf: is the compiler going to compile this away because it's not being assigned to anything? Is an if statement better?

valdrinkoshi: It won't remove it, but I can change it back to an `if` statement if you think is more readable ;)
https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F%252F%2520ADD%2520YOUR%2520CODE%2520HERE%250Afunction%2520updateObj(obj%252C%2520name)%2520%257B%250A%2520%2520obj.name%2520%2526%2526%2520alert('old%2520name%2520'%2520%252B%2520obj.name)%253B%250A%2520%2520obj.name%2520%253D%2520name%253B%250A%2520%2520obj.name%2520%2526%2526%2520alert('new%2520name%2520'%2520%252B%2520obj.name)%253B%250A%257D%250AupdateObj(%257B%257D%252C'New%2520user')%253B%250A%250A

iron-overlay-behavior.html
```diff
@@ -343,16 +336,23 @@
 
       this.__isAnimating = true;
 
-      // requestAnimationFrame for non-blocking rendering
-      this.__openChangedAsync = window.requestAnimationFrame(function() {
-        this.__openChangedAsync = null;
-        if (this.opened) {
-          this._manager.addOverlay(this);
+      // Use requestAnimationFrame for non-blocking rendering.
+      this.__openedChangedRAF && window.cancelAnimationFrame(this.__openedChangedRAF);
```
notwaldorf: same question as above

iron-overlay-behavior.html
notwaldorf: I prefer `if` statements, but I won't object to this if the compiler is happy and warning free :)

notwaldorf:
I think this LGTM? I can't 100% keep all the changes in my mind, but we chatted offline about most of the changes, which sound good in spirit, and the tests look fine, so here's a bear 🐻 .

PR #186, opened 2016-06-02
Fix possible style leaks

naoak:
Fixes Polymer/polymer/issues/3665


valdrinkoshi:
LGTM 👍

notwaldorf:
@valdrinkoshi This should have also updated the `polymer` dependency to `1.1.0` in `bower.json`

valdrinkoshi:
@notwaldorf right, because `v1.1.0` has this [commit](https://github.com/Polymer/polymer/commit/3734c4d), right? I'll prepare a PR for this.

PR #187, opened 2016-06-02
use polymer v1.1.0

valdrinkoshi:
Use `polymer v1.1.0`, which contains [the commit](https://github.com/Polymer/polymer/commit/3734c4d) that ensures the support of styles inside a `template` (see `iron-overlay-backdrop`). This should have been done in #186


notwaldorf:
LGTM

PR #193, opened 2016-06-20
sprinkle some @suppress {missingProperties} for closure compilation

danbeam:
/cc @rictic

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -390,9 +390,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @private
      */
     _shouldBeBehindOverlay: function(overlay1, overlay2) {
-      var o1 = /** @type {?} */ (overlay1);
-      var o2 = /** @type {?} */ (overlay2);
-      return !o1.alwaysOnTop && o2.alwaysOnTop;
+      return !overlay1.alwaysOnTop && overlay2.alwaysOnTop;
     }
   };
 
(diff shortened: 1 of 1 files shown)


rictic:
LGTM but deferring to Valdrin

iron-overlay-manager.html
```diff
@@ -385,6 +386,7 @@
      * @param {!Element} overlay1
      * @param {!Element} overlay2
      * @return {boolean}
+     * @suppress {missingProperties}
```
valdrinkoshi: In the method we're already doing `/** @type {?} */ (overlay1)` for this reason, so we shouldn't need the `@suppress` here, right?

valdrinkoshi:
@danbeam left a comment

iron-overlay-manager.html
danbeam: if it was working, why would I need to send you a pull request? :wink: 

with the cast, chrome's code still fails to compile like this:
```
## /chromium/src/third_party/polymer/v1_0/components-chromium/iron-overlay-behavior/iron-overlay-manager-extracted.js:377: ERROR - Property alwaysOnTop never defined on o1
##       return !o1.alwaysOnTop && o2.alwaysOnTop;
##                  ^
```
with the suppression: compilation succeeds.

obviously somebody added this to fix their error, but Chrome likely has slightly different flag combo.

tl;dr - i can remove the cast if you'd like, but I don't understand who/how it's helping so didn't want to change it.

valdrinkoshi: Hehe yeah I guessed this is still creating problems, was interested to know more context, especially because I added that cast to specifically target that same error noticed in `source: hooli` 😁
May I ask you to remove the cast from this method please? 🙏

danbeam: Done. (removed cast)

valdrinkoshi:
LGTM  :+1:
Thanks @danbeam!

PR #198, opened 2016-06-22
exclude hidden or no-display elements from focusables

### LINKED ISSUE #196: Setting hidden to true can break accessibility in paper-dialog
_From @cbelden on June 15, 2016 18:42_

Came across an issue where a hidden `paper-input` elements prevent correct tabbing behavior in a paper-dialog.

``` html
    <paper-dialog with-backdrop>
      <h2 class="header">Hello</h2>
      <paper-input hidden="true" type="text"></paper-input>
      <div>
        <paper-button dialog-dismiss>Cancel</paper-button>
        <paper-button dialog-confirm>OK</paper-button>
      </div>
    </paper-dialog>
```

After opening the dialog, and clicking on the header, pressing tab does not focus the buttons. Here's the jsfiddle: https://jsfiddle.net/cbelden/tjrt692z/2/

Is this expected behavior? Should we enclose dynamic input elements in `dom-if`s instead of setting the `hidden` attribute?

Note: setting `disabled="true"` on the `paper-input` element had no effect.

_Copied from original issue: Polymer/paper-input#200_

valdrinkoshi:
Fixes #196 by filtering out focusable elements that have `visibility: hidden` or `display: none`, which makes them non-tabbable.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -187,45 +187,50 @@ context. You should place this element as a child of `<body>` whenever possible.
      * @protected
      */
     get _focusableNodes() {
-      // Elements that can be focused even if they have [disabled] attribute.
-      var FOCUSABLE_WITH_DISABLED = [
-        'a[href]',
-        'area[href]',
-        'iframe',
-        '[tabindex]',
-        '[contentEditable=true]'
-      ];
+      var result = [];
+      // If root has non-negative tabindex, it should be the first of the list.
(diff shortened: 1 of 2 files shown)


iron-overlay-behavior.html
```diff
@@ -187,44 +187,51 @@
      * @protected
      */
     get _focusableNodes() {
```
cdata: 😵

valdrinkoshi:
Close this, as it is handled in https://github.com/PolymerElements/iron-overlay-behavior/pull/200

Issue #184, opened 2016-05-19
Dynamically enabled items aren't included in tab order for modal overlays


dpaul:
### Description
If I have a paper-button on an overlay that starts out disabled but becomes enabled while the overlay is shown, that button cannot be tabbed to. This was working in version 1.7.2, commit 0680975e5ec1964e0db9752d19d8a4280c6e9907 seems the likely culprit

### Expected outcome

I should be able to tab to the item.

### Actual outcome

I can't tab to the item.

### Live Demo
http://jsbin.com/foxosogusu/2/edit?html,console,output

### Steps to reproduce
See above.

### Browsers Affected
chrome

valdrinkoshi:
Also with `v1.7.2`, after enabling the Ok button, the focus wouldn't wrap http://jsbin.com/zakuni/1/edit?html,output

In `v.1.7.2`, `iron-overlay-behavior` would search for the focusable elements when its children changed (before the `paper-button` would be disabled). In the latest version, this search is done when the overlay is fully opened (right before sending `iron-overlay-opened`).

There is already an effort going on to separate the focus wrapping/trapping into a separate behavior (see #179), and in PR #178 I'm updating `iron-overlay-behavior` to use this new behavior. Once merged, you'll be able to set the first/last focusable like this: http://jsbin.com/zanene/edit?html,output

freshp86:
cc: @danbeam 

I am facing the same problem in Chrome's Polymer settings (see https://bugs.chromium.org/p/chromium/issues/detail?id=623907#c3). Specifically we have a dialog with a "submit" button that is disabled initially, and it only gets enabled after the user fills in some text inputs. The button is not reachable via TAB key. See example at https://jsfiddle.net/944v704y/1/. Probably a regression from https://github.com/PolymerElements/iron-overlay-behavior/commit/1216db42b441e2dada1e08fefe8d3bad45796401.

valdrinkoshi:
This has been fixed by introducing a new method, [invalidateTabbables()](https://www.webcomponents.org/element/PolymerElements/iron-overlay-behavior/behaviors/Polymer.IronOverlayBehaviorImpl#method-invalidateTabbables) to invalidate the tabbable elements. You'll have to call `overlay.invalidateTabbables()` right after the state of your focusable elements changes.

PR #199, opened 2016-06-22
remove caching of __first/lastFocusableNodes

valdrinkoshi:
Fixes https://github.com/PolymerElements/iron-overlay-behavior/issues/184 by removing the caching of `__firstFocusableNode, __lastFocusableNode`.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -234,8 +234,6 @@ context. You should place this element as a child of `<body>` whenever possible.
       // with-backdrop needs tabindex to be set in order to trap the focus.
       // If it is not set, IronOverlayBehavior will set it, and remove it if with-backdrop = false.
       this.__shouldRemoveTabIndex = false;
-      // Used for wrapping the focus on TAB / Shift+TAB.
-      this.__firstFocusableNode = this.__lastFocusableNode = null;
       // Used for requestAnimationFrame when opened changes.
       this.__openChangedAsync = null;
       // Used for requestAnimationFrame when iron-resize is fired.
@@ -413,11 +411,6 @@ context. You should place this element as a child of `<body>` whenever possible.
       this.notifyResize();
       this.__isAnimating = false;
 
-      // Store it so we don't query too much.
-      var focusableNodes = this._focusableNodes;
-      this.__firstFocusableNode = focusableNodes[0];
-      this.__lastFocusableNode = focusableNodes[focusableNodes.length - 1];
-
       this.fire('iron-overlay-opened');
     },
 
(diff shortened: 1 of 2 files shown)


iron-overlay-behavior.html
```diff
@@ -413,11 +411,6 @@
       this.notifyResize();
       this.__isAnimating = false;
 
-      // Store it so we don't query too much.
```
notwaldorf: Have you done any perf measurements to see how this compares to not caching it?

notwaldorf:
This LGTM in principle, but I would love to see some perf numbers -- how much it cost to tab before, how much it costs now (just in case re-calculating all the focusable nodes on a sizeable, deep dialog is actually awful)

valdrinkoshi:
@notwaldorf I did some tests like holding the TAB key for 2-3 sec, then Shift+Tab keys for 2-3 sec http://jsbin.com/yicewuy/1/edit?html,output
The difference between what's on master and in this PR is around 4ms.

Master:

screen shot 2016-07-15 at 1 45 33 am


PR (see `_focusableNodes`):

screen shot 2016-07-15 at 1 43 27 am


Should I make more complex scenarios and see how bad it is? It is strictly dependent on the speed of `_focusableNodes`, which btw can be overridden to return only first and last focusables.

valdrinkoshi:
Here a example with more tabbable elements, logging the time of `_focusableNodes` with `console.time` and `console.timeEnd`; the avg is around 3-4ms:
tab-heavy

notwaldorf:
@valdrinkoshi Looking at the comments this looks like I already LGTMed it, so you'll probably need  a rebase?

Re: the perf notes: I think I am more worried about something that's very deep (like a bunch of nested elements or divs). I guess 3ms isn't that bad

valdrinkoshi:
I've updated the approach with which tabbables can be updated by introducing `invalidateTabbables()`. The performance is improved in comparison to master, because the first/last focusable nodes are calculated only once when the user presses tab/shift+tab, and then these are cached (and reused on the following opening of the overlay too). We invalidate these values if a node is added/removed, and for all other cases (e.g. attribute `disabled` changes on a button), we require the user to call `invalidateTabbables()`

@notwaldorf @cdata PTAL, WDYT?

notwaldorf:
Still LGTM i think (the diff is way smaller now from what i remember) :)

valdrinkoshi:
Tests are already failing on master with the latest `web-component-tester`, e.g. see
v4.3.4 https://travis-ci.org/PolymerElements/iron-overlay-behavior/builds/149623699 and v4.3.1 https://travis-ci.org/PolymerElements/iron-overlay-behavior/builds/149624197

Tested this locally, will merge.

PR #200, opened 2016-06-28
add iron-focusables-helper

### LINKED ISSUE #192: Focus should work in shadow DOM in addition to light DOM
### Description
When an element with iron overlay behavior contains custom elements, the overlay with backdrop should be able to tab into the shadow DOM of those elements. If not a full fix, would be nice to be able to provide focusable nodes to the containing element and have a behavior that finds the focusable nodes in an element.

### Expected outcome
The elements inside of shadow DOM should be tabbed to.

### Actual outcome

Only elements in light DOM can be tabbed to (captured by focusableNodes)

### Steps to reproduce
1. Use an element with iron overlay behavior and withBackdrop on. 
2. Create a custom element and place it inside of the overlay.
3. Try to tab through element

### LINKED ISSUE #194: Distributed children break focus wrapping
### Description
Focusable children that are distributed into an overlay by another element break the focus wrapping
```html
<template>
  <p> x-element </p>
  <paper-dialog>
    <content></content>
  </paper-dialog>
</template>
...

<x-element>
  <button> Breaks focus wrapping </button>
</x-element>
```

### Live Demo
 http://jsbin.com/tojoroyibi/1/edit?html,output

### LINKED ISSUE #196: Setting hidden to true can break accessibility in paper-dialog

### LINKED ISSUE #209: Order of focusable elements gets messed up by tabIndex sorting
### Description
`get _focusableNodes()` uses `Array.sort` to sort focusable elements by `tabIndex` ([see here](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L228)). However, `Array.sort` is not guaranteed to be stable ([see documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)). This leads to the order of focusable elements getting messed up and makes `__first/__lastFocusableNodes` point to incorrect nodes which results in unexpected behavior when using the tab key to sequentially walk through focusable nodes.

E.g. in Chrome it can be reproduced once the array has more than 10 elements: https://jsbin.com/rurowalosa/edit?js,console

valdrinkoshi:
Fixes #196, fixes #194, fixes #192.
`iron-focusables-helper` provides `getTabbableNodes` method, used to compute the tabbable elements within a specific node, searching both in the light and shadow dom, sorting by tabindex.
`_focusableNodes` has been changed to invoke this method. Tests have been updated to look for the focusable nodes once the overlays are opened (as it is required for a node to be visible in order to be considered tabbable).
Sorting by tabindex is a stable sort ([merge sort](https://www.youtube.com/watch?v=XaqR3G_NVoo)), since `Array.prototype.sort` is not guaranteed to be stable (fixes #209).

### DIFF
--- a/iron-focusables-helper.html
+++ b/iron-focusables-helper.html
@@ -1,0 +1,135 @@
+<!doctype html>
+<!--
+@license
+Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="../polymer/polymer.html">
+
+<script>
+  (function() {
+    'use strict';
+
+    Polymer.IronFocusablesHelper = {
+
+      /**
+       * Returns a sorted array of tabbable nodes, including the root node.
+       * It searches the tabbable nodes in the light and shadow dom of the chidren,
+       * sorting the result by tabindex.
+       * @param {!Node} node
+       * @return {Array<Node>}
+       */
+      getTabbableNodes: function(node) {
+        var result = [];
+        // If there is at least one element with tabindex > 0, we need to sort
+        // the final array by tabindex.
+        var maxTabIndex = this._findTabbableNodes(node, result);
+        maxTabIndex > 0 && result.sort(this._tabIndexSort);
+        return result;
+      },
+
+      /**
+       * Returns if a node is focusable.
+       * @param {!Node} node
+       * @return {boolean}
+       */
+      isFocusable: function(node) {
+        // shadow roots won't have this method.
+        if (!node.hasAttribute) {
+          return false;
+        }
+        var name = node.localName;
+        // Elements that can be focused even if they have [disabled] attribute.
+        if (name === 'iframe' ||
+          ((name === 'a' || name === 'area') && node.hasAttribute('href')) ||
+          node.hasAttribute('tabindex') || node.hasAttribute('contentEditable')) {
+          return true;
+        }
+        // Elements that cannot be focused if they have [disabled] attribute.
+        return /^(input|select|textarea|button|object)$/.test(name) && !node.disabled;
+      },
+
+      /**
+       * Returns if a node is tabbable.
+       * @param {!Node} node
+       * @return {boolean}
+       */
+      isTabbable: function(node) {
+        return node.tabIndex >= 0 && this.isFocusable(node) && this._isVisible(node);
+      },
+
+      /**
+       * Searches for nodes that are tabbable and adds them to the `result` array.
+       * Returns the maximum tabindex of the added nodes.
+       * @param {!Node} node The starting point for the search; it's added if tabbable.
+       * @param {!Array<Node>} result
+       * @return {Number}
+       * @private
+       */
+      _findTabbableNodes: function(node, result) {
+        var maxTabIndex = 0;
+        // Skip #text nodes. If not visible, no need to explore children.
+        if (node.nodeType !== 3 && this._isVisible(node)) {
+          if (this.isTabbable(node)) {
+            result.push(node);
+            maxTabIndex = node.tabIndex;
+          }
+          var children;
+          if (node.localName === 'content') {
+            children = Polymer.dom(node).getDistributedNodes();
+          } else {
+            // Use shadow root if possible, will check for distributed nodes.
+            children = Polymer.dom(node.root || node).children;
+          }
+          for (var i = 0; i < children.length; i++) {
+            maxTabIndex = Math.max(this._findTabbableNodes(children[i], result), maxTabIndex);
+          }
+        }
+        return maxTabIndex;
+      },
+
+      /**
+       * Returns false if the node has `visibility: hidden` or `display: none`
+       * @param {!Node} node
+       * @return {boolean}
+       * @private
+       */
+      _isVisible: function(node) {
+        // Check inline style first, might save a re-flow.
+        return this._isStyleVisible(node.style) && this._isStyleVisible(window.getComputedStyle(node));
+      },
+
+      /**
+       * Returns false if the style is `visibility: hidden` or `display: none`
+       * @param {!CSSStyleDeclaration} style
+       * @return {boolean}
+       * @private
+       */
+      _isStyleVisible: function(style) {
+        return (style.visibility !== 'hidden' && style.display !== 'none');
+      },
+
+      /**
+       * Sort by tabindex. Move elements with tabindex = 0 to the end to match
+       * browser behavior, e.g. [0, 3, 1, 2] --> [1, 2, 3, 0]
+       * @param {!Node} a
+       * @param {!Node} b
+       * @return {Number}
+       * @private
+       */
+      _tabIndexSort: function(a, b) {
+        // Move tabindex = 0 elements to the end of the list.
+        if (a.tabIndex === 0 || b.tabIndex === 0) {
+          // The one with bigger `tabindex` goes on top of the list.
+          return b.tabIndex - a.tabIndex;
+        }
+        // The one with smaller `tabindex` goes on top of the list.
+        return a.tabIndex - b.tabIndex;
+      }
+    };
+  })();
+</script>
--- a/iron-focusables-helper.html
+++ b/iron-focusables-helper.html
@@ -70,7 +70,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @return {boolean}
        */
       isTabbable: function(node) {
-        return node.tabIndex >= 0 && this.isFocusable(node) && this._isVisible(node);
+        // Check for the attribute "tabindex" instead of the element property
+        // `tabIndex` since browsers assign different values to it e.g.
+        // <div contenteditable></div> has tabIndex = -1 in FF, tabIndex = 0 in
+        // other browsers.
+        return this.isFocusable(node) && this._isVisible(node) &&
+          node.getAttribute('tabindex') !== '-1';
       },
 
       /**
--- a/iron-focusables-helper.html
+++ b/iron-focusables-helper.html
@@ -38,19 +38,30 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @return {boolean}
        */
       isFocusable: function(node) {
+        // From http://stackoverflow.com/a/1600194/4228703:
+        // There isn't a definite list, it's up to the browser. The only
+        // standard we have is DOM Level 2 HTML https://www.w3.org/TR/DOM-Level-2-HTML/html.html,
+        // according to which the only elements that have a focus() method are
+        // HTMLInputElement,  HTMLSelectElement, HTMLTextAreaElement and
+        // HTMLAnchorElement. This notably omits HTMLButtonElement and
+        // HTMLAreaElement.
+        // Referring to these tests with focusables in different browsers
+        // http://allyjs.io/data-tables/focusable.html
+
         // shadow roots won't have this method.
         if (!node.hasAttribute) {
           return false;
         }
         var name = node.localName;
-        // Elements that can be focused even if they have [disabled] attribute.
-        if (name === 'iframe' ||
-          ((name === 'a' || name === 'area') && node.hasAttribute('href')) ||
-          node.hasAttribute('tabindex') || node.hasAttribute('contentEditable')) {
-          return true;
+        // These elements cannot be focused if they have [disabled] attribute.
+        if (/^(input|select|textarea|button|object)$/.test(name)) {
+          return !node.disabled;
         }
-        // Elements that cannot be focused if they have [disabled] attribute.
-        return /^(input|select|textarea|button|object)$/.test(name) && !node.disabled;
+        // These elements can be focused even if they have [disabled] attribute.
+        return name === 'iframe' ||
+          (/^(a|area)$/.test(name) && node.hasAttribute('href')) ||
+          node.hasAttribute('tabindex') ||
+          node.hasAttribute('contenteditable');
       },
 
       /**
@@ -99,18 +110,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @private
        */
       _isVisible: function(node) {
-        // Check inline style first, might save a re-flow.
-        return this._isStyleVisible(node.style) && this._isStyleVisible(window.getComputedStyle(node));
-      },
-
-      /**
-       * Returns false if the style is `visibility: hidden` or `display: none`
-       * @param {!CSSStyleDeclaration} style
-       * @return {boolean}
-       * @private
-       */
-      _isStyleVisible: function(style) {
-        return (style.visibility !== 'hidden' && style.display !== 'none');
+        // Check inline style first to save a re-flow. If looks good, check also
+        // computed style.
+        if (node.style.visibility !== 'hidden' && node.style.display !== 'none') {
+          var style = window.getComputedStyle(node);
+          return (style.visibility !== 'hidden' && style.display !== 'none');
+        }
+        return false;
       },
 
       /**
--- a/iron-focusables-helper.html
+++ b/iron-focusables-helper.html
@@ -27,8 +27,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var result = [];
         // If there is at least one element with tabindex > 0, we need to sort
         // the final array by tabindex.
-        var maxTabIndex = this._findTabbableNodes(node, result);
-        maxTabIndex > 0 && result.sort(this._tabIndexSort);
+        var needsSortByTabIndex = this._collectTabbableNodes(node, result);
+        if (needsSortByTabIndex) {
+          result.sort(this._tabIndexSort);
+        }
         return result;
       },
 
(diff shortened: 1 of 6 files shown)


iron-focusables-helper.html
```diff
@@ -0,0 +1,135 @@
...
+       * sorting the result by tabindex.
+       * @param {!Node} node
+       * @return {Array<Node>}
+       */
+      getTabbableNodes: function(node) {
+        var result = [];
+        // If there is at least one element with tabindex > 0, we need to sort
+        // the final array by tabindex.
+        var maxTabIndex = this._findTabbableNodes(node, result);
+        maxTabIndex > 0 && result.sort(this._tabIndexSort);
```
bicknellr: `Array.prototype.sort` isn't guaranteed to be stable.

valdrinkoshi: Uhm, in which way/which browsers? I've written unit tests that check for the result of this method & if it returns a correctly sorted array

bicknellr: Practically speaking, I doubt any browsers use an unstable sort, but I figured it might be worth pointing out.

iron-focusables-helper.html
```diff
@@ -0,0 +1,135 @@
...
+      },
+
+      /**
+       * Returns false if the style is `visibility: hidden` or `display: none`
+       * @param {!CSSStyleDeclaration} style
+       * @return {boolean}
+       * @private
+       */
+      _isStyleVisible: function(style) {
+        return (style.visibility !== 'hidden' && style.display !== 'none');
```
bicknellr: I think you'll need to test if `getBoundingClientRect` returns all zeroes like you mentioned before, rather than checking for `display: none;`: http://jsbin.com/moyoma/1/edit?html,output

valdrinkoshi: The example you shared looks odd, doesn't contain `getBoundingClientRect` :p
Anyways `visibility: hidden` and `display: none` are the only 2 css properties that affect the tab order. Also, this check will affect the search too, if it should check the node children as well (see line 76); checking for the bounding rect would be wrong in a case like this http://jsbin.com/tepuwuq/1/edit?html,console,output (check the log)

iron-focusables-helper.html
valdrinkoshi: I couldn't experience any issue so far, and the tests ensured it works as expected. Will keep an eye on it if it shows a weird behavior!

iron-focusables-helper.html
bicknellr: Yeah, that comment was poorly worded, I was just trying to show the problem. I didn't consider nested `position: absolute` but it looks like line 76 avoids the issue from the example completely!

iron-focusables-helper.html
```diff
@@ -0,0 +1,135 @@
...
+
+      /**
+       * Searches for nodes that are tabbable and adds them to the `result` array.
+       * Returns the maximum tabindex of the added nodes.
+       * @param {!Node} node The starting point for the search; it's added if tabbable.
+       * @param {!Array<Node>} result
+       * @return {Number}
+       * @private
+       */
+      _findTabbableNodes: function(node, result) {
```
bicknellr: The signature of this function feels very specific to its use in `getTabbableNodes`; given the name `_findTabbableNodes`, I'd expect this to return the list rather than taking a list to modify. You'd be able to take lines 88-90 out of this function too if you computed `maxTabIndex` in `getTabbableNodes`.

iron-focusables-helper.html
```diff
@@ -0,0 +1,135 @@
...
+       * @return {boolean}
+       */
+      isFocusable: function(node) {
+        // shadow roots won't have this method.
+        if (!node.hasAttribute) {
+          return false;
+        }
+        var name = node.localName;
+        // Elements that can be focused even if they have [disabled] attribute.
+        if (name === 'iframe' ||
```
bicknellr: Is there a spec you could include a link to that lists / explains these?

iron-focusables-helper.html
valdrinkoshi: The reasoning for this layout is to avoid too many searches & achieve recursion, as we need to do a depth first search in order to keep the tabindex order. The fact that we return the `maxTabIndex` is just to support cases where we might have a tabindex > 0 (which is an anti-pattern, but I guess we have to support it, right?).
I could move the searching of the max tabindex outside, but that would require another for loop over the result :/

iron-focusables-helper.html
```diff
@@ -0,0 +1,135 @@
...
+
+      /**
+       * Returns false if the node has `visibility: hidden` or `display: none`
+       * @param {!Node} node
+       * @return {boolean}
+       * @private
+       */
+      _isVisible: function(node) {
+        // Check inline style first, might save a re-flow.
+        return this._isStyleVisible(node.style) && this._isStyleVisible(window.getComputedStyle(node));
```
bicknellr: `_isStyleVisible` is only used here and is small enough that it might be worth inlining it.

valdrinkoshi: 👍

iron-focusables-helper.html
bicknellr: Either way, the tab index of each element in the list has to be reviewed. As written now, the check is interleaved with the traversal; splitting it out won't change the time complexity but would let this function do one less thing. If there was some sort of early return behavior that could be exploited here, I'd agree, but I don't think there is.

iron-focusables-helper.html
valdrinkoshi: Unfortunately there is not a definite list, it's up to the browsers (lol). I should indeed put some comments, and maybe links too:
http://stackoverflow.com/a/1600194/4228703
http://allyjs.io/data-tables/focusable.html

bicknellr: Oh wow, that table is huge!

iron-focusables-helper.html
valdrinkoshi: Yes, the tab index of each element has to be reviewed. Currently, I do that on the same loop where i collect the tabbables. This allows me to save another loop over the tabbables after they get collected.
```javascript
// o(node-depth) search
var tabbables = [];
var maxTabIndex = this._findTabbables(node, tabbables);

// o(node-depth) + o(tabbables) search
var tabbables = this._findTabbables(node);
var maxTabIndex = this._findMaxTabIndex(tabbables);
```
The second option is more readable but less performant. How to proceed on this one?

iron-focusables-helper.html
valdrinkoshi: Yeah 😄🔫 
I've updated the comments ;) Checking why it fails on firefox..!

valdrinkoshi: Found out the issue & fixed (in the latest commit)

iron-focusables-helper.html
bicknellr: All the work done in `_findMaxTabIndex` still happens in the original `_findTabbables`. I think the second case can even potentially be more performant: considering only work done to find the max tab index, the original `_findTabbables` has best case O(n) because line 105 occurs for every node traversed, even when the nodes in question are not in the tab order. However, splitting this out into a separate loop gives a best case of O(1) because the list, which will be some subset of the traversed elements, can potentially be empty. (Both still have worst case O(n).)

valdrinkoshi: Just to recap:
- we don't want to sort if all elements have `tabindex = 0` -> look if there is one element with `tabIndex > 0`
- we have to traverse all the children to find the tabbables -> minimum is `O(n)`
- to filter the tabbable elements, we must look at their `tabindex` to be `>= 0`

Let's compare the two options in some scenarios for `n` nodes of which `m` are tabbable, and `i` is the index of the tabbable with `tabindex > 0`:
```javascript
// option 1
var tabbables = [];
var maxTabIndex = this._findTabbables(node, tabbables);

// option 2
var tabbables = this._findTabbables(node);
var maxTabIndex = this._findMaxTabIndex(tabbables);
```
#### scenario 1 (worst case scenario)
 `m = n` (all are tabbable), `i = -1` (no tabbable with `tabindex > 0`)
- option 1: `O(n)`
- option 2: `O(n) + O(n)` (needs to search until the end of tabbables)

#### scenario 2
`m = n/10` (only 10% tabbable), `i = -1` (no tabbable with `tabindex > 0`)
- option 1: `O(n)`
- option 2: `O(n) + O(n/10)` (needs to search until the end of tabbables)

#### scenario 3
`m = n/10` (only 10% tabbable), `i = m/2` (the tabbable on the middle has `tabindex > 0`)
- option 1: `O(n)`
- option 2: `O(n) + O(n/20)` (can stop as soon as it finds the tabbable with `tabindex > 0`)

#### scenario 4 (best case scenario)
`m = 0` (no tabbables) 
- option 1: `O(n)`
- option 2: `O(n)`

valdrinkoshi: If we average of worst and best scenarios, we see that on average option 2 is 50% slower than option 1.
In practice probably this doesn't make a huge difference, I just would like to understand if you see a better, more performant way to tackle this. Please correct me if I wrote something wrong :)

bicknellr: I'm pretty sure all of those are O(n) since "O(n) + O(n)" = O(n) but, expanding out the two pieces of work (gather and max), option 1 would be "O(n) + O(n)" (maybe more clearly written as something like "n * 2O(1)" in this case) for all of those situations, even though it uses a single loop.

```javascript
function sumAndConcat1(list) {
  var sum = 0;
  var string = "";
  for (var i = 0; i < list.length; i++) {
    sum += list[i];
    string += list[i];
  }
  return {sum, string};
}


function sumAndConcat2(list) {
  var sum = 0;
  var string = "";
  for (var i = 0; i < list.length; i++) {
    sum += list[i];
  }
  for (var i = 0; i < list.length; i++) {
    string += list[i];
  }
  return {sum, string};
}
```

I don't think the use of a single for loop in `sumAndConcat1` shouldn't give it any performance advantage over `sumAndConcat2`. And, if the second loop in `sumAndConcat2` could perform fewer than `list.length` iterations (assuming the function was doing something else), wouldn't that be an advantage for `sumAndConcat2`?

However, I suppose the end to my argument wouldn't have much benefit anyway given the operation being optimized out (max) has such a small cost. :P I still think the function's signature is weird but, yeah, the rest of my argument is more pedantic than practical.

valdrinkoshi: Since we only need to know if the maxTabIndex is greater than zero, and not its actual value, I could remove the `Math.max` part and make it return a boolean. I'll check also a better/clearer signature for this method 👍

iron-focusables-helper.html
valdrinkoshi: @bicknellr you were right, the issue was reported here https://github.com/PolymerElements/iron-overlay-behavior/issues/209
Will fix this!

kornalius:
Got a bug in iron-focusables-helper in _isVisible.

`if (node.style.visibility !== 'hidden' && node.style.display !== 'none') {`
"Cannot read property 'visibility' of undefined"

node.style does not exist on #comment nodes.

valdrinkoshi:
Good catch @kornalius, fixed in the last commit!

valdrinkoshi:
Did some tests with `heavy.html` (see below), and on average:
- `_onCaptureTab` when no wrapping is ~ 0.02ms
- `_onCaptureTab` when wrapping is ~1ms

Querying for tabbable nodes is the part that takes the most of the time:
- implementation on master (doesn't sort correctly with this example) takes **~0.8ms**
- the implementation of this PR does sort correctly but takes **~5ms** (I suspect the implementation of the merge sort is the thing slowing it down)

> heavy.html

```html

<button onclick="plain.open()">overlay</button>
<simple-overlay with-backdrop id="plain">
  <a href="#">click</a>
  <a href="#">click</a>
  <a href="#">click</a>
  <input>
  <paper-input></paper-input>
  <paper-input></paper-input>
  <paper-input></paper-input>
  <paper-input></paper-input>
  <paper-input></paper-input>
  <div contenteditable>contenteditable</div>
  <div tabindex="0">tabindex 0</div>

  <a href="#">click</a>


  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <paper-dropdown-menu>
    <ul class="dropdown-content">
      <button>one</button>
      <button>two</button>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
      <li><a href="#">click</a></li>
    </ul>
  </paper-dropdown-menu>
  <button onclick="plain.close()">Close</button>
</simple-overlay>

**Comment by valdrinkoshi** [comment]
@cdata FYI ^

**Comment by notwaldorf** [comment]
going to unassign myself from this since I don't seem needed 😂

File: iron-focusables-helper.html [hunk 6]
```diff
@@ -0,0 +1,207 @@
...
+      /**
+       * Merge sort iterator, merges the two arrays into one, sorted by tab index.
+       * @param {!Array<HTMLElement>} left
+       * @param {!Array<HTMLElement>} right
+       * @return {Array<HTMLElement>}
+       * @private
+       */
+      _mergeSortByTabIndex: function(left, right) {
+        var result = [];
+        while ((left.length > 0) && (right.length > 0)) {
```
bicknellr: I was totally about to leave comments about how everything's going to break if either is empty and the concat with 'empty' arrays until I noticed that this is `&&` and not `||`!

iron-focusables-helper.html
```diff
@@ -0,0 +1,207 @@
...
+      /**
+       * Searches for nodes that are tabbable and adds them to the `result` array.
+       * Returns if the `result` array needs to be sorted by tabindex.
+       * @param {!Node} node The starting point for the search; added to `result`
+       * if tabbable.
+       * @param {!Array<HTMLElement>} result
+       * @return {boolean}
+       * @private
+       */
+      _collectTabbableNodes: function(node, result) {
```
bicknellr: I don't think this method accounts for focus navigation scopes. Here, all the nodes are sorted at the end but each scope needs to be sorted separately. For example:
```html
<div id="root">
  <!-- shadow -->
    <slot name="a">
    <slot name="b">
  <!-- /shadow -->
  <input id="A" slot="a">
  <input id="B" slot="b">
  <input id="C" slot="b" tabindex="1">
</div>
```
I think calling `getTabbableNodes(#root)` would return [`#C`, `#A`, `#B`], but it should return [`#A`, `#C`, `#B`] because `slot[name=b]` creates a focus navigation scope.

valdrinkoshi: That's true, but I wouldn't encourage the support of `tabindex > 0` for distributed content (I woudn't encourage its usage at all). The current implementation offers a solution that works in 80% of the cases, mainly for shady dom. But there are so many messed up permutations (e.g. ShadowDom v1 + delegatesFocus = false) where it is not clear yet what's the correct, desired behavior for tab order 😕

bicknellr: Something I didn't think about yesterday: if this doesn't intend to support tabindex, what's the point of the sort at all? If everything is assumed to have `tabindex` 0, document order *is* sequential focus navigation order.

bicknellr: That wasn't particularly well worded.. What I mean is, if calling `getTabbableNodes` with an ancestor of an element with `tabindex > 0` is *expected* to break, then why attempt to sort? Why not just filter them out?

valdrinkoshi: Uhm, tested your example and indeed slots will scope the focus within their content..even without tabindex sorting this will be a problem for v1. So I should solve it. Will come up with a solution for this ;)

mercmobily:
Thanks for your work on this @valdrinkoshi . This is the kind of contribution I feel I will never be able to make. It would be great to get a proper solution to https://github.com/PolymerElements/paper-dialog/issues/125 !

romancm:
@valdrinkoshi - Is there any update on this PR? I'm interested in having this PR merged in. Please let me know if there's anything I can do to help. Thanks!

iron-focusables-helper.html
bicknellr: (from offline discussion) As this isn't part of the Polymer 2.0 upgrade yet and Polymer prior to 2 only uses Shadow DOM v0, there's no need for this to actually cover v1 focus behavior. `<slot>` to `<content>` conversion in Polymer 1.7 means that Shadow DOM v1-like markup ends up as v0 anyway. However, this means handling Shadow DOM v1 focus behavior will be part of porting to hybrid / 2.0. As long as you're ok with that, ignoring v1 here SGTM.

valdrinkoshi: yep, I've added a comment for the future maintainer at line 116 👌

iron-focusables-helper.html
```diff
@@ -0,0 +1,220 @@
...
+          } else {
+            result.push(left.shift());
+          }
+        }
+
+        return result.concat(left, right);
+      },
+
+      /**
+       * Returns if an element has lower tab order compared to another element
```
bicknellr: Which is `a` and which is `b`?

valdrinkoshi:
@bicknellr good catch, clarified the jsdoc of `_hasLowerTabOrder` 👌

PR #202, opened 2016-06-30
move focus a tick after opened changes

### LINKED ISSUE #201: Long delay between paper-dropdown-menu open and focus
_From @rictic on April 19, 2016 18:49_

Moved from internal bug:

See: https://elements.polymer-project.org/elements/paper-dropdown-menu?view=demo:demo/index.html&active=paper-dropdown-menu

In paper-dropdown-menu I'm seeing a 540ms delay between when the dropdown is opened and the first item gets rendered as focused. This creates a jarring change after the UI should already be stable.

This is caused by focus being applied in `iron-overlay-behavior`'s `_finishRenderOpened`. If focus causes visual changes then it should be applied before the open animation.

Alternatively the style for the focus should be much less glaring.

```css
paper-menu {
  --paper-menu-focused-item: {
    display: none;
  };
}
```

_Copied from original issue: PolymerElements/paper-dropdown-menu#150_

valdrinkoshi:
Fixes #201 by applying the focus a tick after `opened` changes. ~~Avoid calling `blur` when an overlay gets closed, since a closed overlay will be `display: none` hence non-focusable.~~

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -351,6 +351,8 @@ context. You should place this element as a child of `<body>` whenever possible.
           this._prepareRenderOpened();
           this._renderOpened();
         } else {
+          // Move the focus before actually closing.
+          this._applyFocus();
           this._renderClosed();
         }
       }.bind(this));
@@ -387,6 +389,9 @@ context. You should place this element as a child of `<body>` whenever possible.
       this.refit();
       this._finishPositioning();
 
+      // Move the focus to the child node with [autofocus].
+      this._applyFocus();
+
       // Safari will apply the focus to the autofocus element when displayed for the first time,
       // so we blur it. Later, _applyFocus will set the focus if necessary.
       if (this.noAutoFocus && document.activeElement === this._focusNode) {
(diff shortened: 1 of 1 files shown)


iron-overlay-behavior.html
```diff
@@ -477,7 +478,6 @@
           this._focusNode.focus();
         }
       } else {
-        this._focusNode.blur();
```
notwaldorf: Hmm, is this bad? Like, will there be side effects because you used to be blurring this node and you're not anymore?

notwaldorf:
I think this LGTM, but I have a question about the `blur` bit. Isn't it weird you're not calling blur anymore? Or it makes sense because that element isn't even showing at that point?

iron-overlay-behavior.html
valdrinkoshi: The overlay is not displayed anymore at this point (it is `opened = false`, which means `display: none`, hence not focusable).

valdrinkoshi: Uhm, just tested this http://jsbin.com/vapoci/1/edit?html,console,output and it has different behavior on FF compared to Chrome, aka even if `display: none` the previously focused element is still the `document.activeElement`...Will put back this line where it was!

valdrinkoshi:
Feedback incorporated, waiting for the build to be green before merging this

PR #211, opened 2016-08-12
Avoid calling `blur()` when method don't exists

frncesc:
The internal method `_applyFocus` of [iron-overlay-behavior](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L455) makes a call to `this._focusNode.blur();`.

When `this._focusNode` has no method "blur" defined, this causes an exception.

I just discovered that __MS Explorer 11__ and __MS Edge__ don't define any "blur" method on the `<svg>` element, so when used as a button calling `close()` on `iron-overlay` objects the process silently drops, making impossible to close the overlay.

Please check this plunk in Explorer or Edge:
https://plnkr.co/edit/rKwnv4zXFWvP9GVzPeYQ?p=preview

As yo will see, the "close" button works as expected, while the SVG blue cross fails:
(checked on SauceLabs with Windows 10 virtual machines running Explorer 11 and Edge 13)

As a workaround, I propose to modify iron-overlay-behavior, checking if `this._focusNode.blur`
exists before calling it.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -376,7 +376,8 @@ context. You should place this element as a child of `<body>` whenever possible.
       // Safari will apply the focus to the autofocus element when displayed
       // for the first time, so we make sure to return the focus where it was.
       if (this.noAutoFocus && document.activeElement === this._focusNode) {
-        this._focusNode.blur();
+        if (this._focusNode.blur)
+          this._focusNode.blur();
         this.__restoreFocusNode.focus();
       }
     },
@@ -459,7 +460,8 @@ context. You should place this element as a child of `<body>` whenever possible.
         }
       }
       else {
-        this._focusNode.blur();
+        if (this._focusNode.blur)
+          this._focusNode.blur();
         this._focusedChild = null;
         // Restore focus.
         if (this.restoreFocusOnClose && this.__restoreFocusNode) {


notwaldorf:
@valdrinkoshi should we in parallel maybe file a bug with Edge? This sounds a bit bananas :sob:

valdrinkoshi:
I don't think this should be fixed in `iron-overlay-behavior` as it is clearly a bug on Edge (it has this issue for declared svg, if you create it imperatively it works http://jsbin.com/kivufo/2/edit?html,console,output)
I suggest to file a bug with Edge and as a temporary solution have something like 
```javascript
var svgs = document.body.querySelectorAll('svg');
if (svgs.length && (!svgs[0].focus || !svgs[0].blur)) {
  for (var i = 0; i < svgs.length; i++) {
    svgs[i].focus = svgs[i].focus || function() { console.log('lol i dont know how to focus'); };
    svgs[i].blur = svgs[i].blur || function() { console.log('lol i dont know how to blur'); };
  }
}

**Comment by valdrinkoshi** [comment]
Closing this, I'll file the issue on Edge and link it here ;)

**Comment by valdrinkoshi** [comment]
Here the issue I filed https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8479637/

**Comment by frncesc** [comment]
OK,

I'm not a Windows user, just discovered this annoying bug while testing a Polymer web app with different browsers in SauceLabs.

I will implement a specific workaround in my app, based on your solution.

Thank you and congratulations for the excellent work done in Polymer!!

**Comment by valdrinkoshi** [comment]
This polyfill seems to do the trick ;)
```javascript
SVGSVGElement.prototype.focus = SVGSVGElement.prototype.focus || HTMLElement.prototype.focus;
SVGSVGElement.prototype.blur = SVGSVGElement.prototype.blur || HTMLElement.prototype.blur;
```

http://jsbin.com/teqatuz/5/edit?html,output

PR #213, opened 2016-08-18
delegate tab, esc, focus, tab to correct top overlay

### LINKED ISSUE #212: IronOverlayManager should skip elements that have noCancelOnOutsideClick/Esc when a close event is handled
### Description

An overlay with noCancelOnOutsideClick prevents all overlays without cancel to stay open. 
Example:
There are two overlays. A dropdown opens and a user does an action on the dropdown triggering a toast to open. The toast has noCancelOnOutsideClick = true, but the dropdown should close on outside click or esc key. 
### Expected outcome

The dropdown stays open when the user clicks outside of the dropdown for as long as the toast is open.
### Actual outcome

The dropdown should close when the user clicks outside of it, but the toast should remain until it closes (in this case after a few seconds). The manager should skip elements that have noCancel on them, and cancel the next overlay that has noCancel = false.

valdrinkoshi:
Fixes #212 by delegating `click, esc, focus, tab` events to the correct top overlay.
- `click` -> look for top overlay with `noCancelOnOutsideClick = false`
- `esc` -> look for top overlay with `noCancelOnEscKey = false`
- `focus, tab` -> look for top overlay with `withBackdrop = true` (not required by the issue but done for completeness)

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -406,9 +406,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @protected
      */
     _onCaptureClick: function(event) {
-      if (!this.noCancelOnOutsideClick) {
-        this.cancel(event);
-      }
+      this.cancel(event);
     },
 
     /**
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
closing this as it would be a breaking change, and in some cases not the desired behavior (see https://github.com/PolymerElements/iron-overlay-behavior/issues/212#issuecomment-333225393)

PR #214, opened 2016-09-03
remove overwritten _openChanged method

quangv:
correct me if i'm wrong, but it looks like `_openChanged` method was declared twice?

https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L612-L628


quangv:
I signed it!

valdrinkoshi:
Not really, the second one has two _  
We should probably give another name to the private one, instead of __openedChanged it should maybe be __renderOpenedChanged

PR #219, opened 2016-09-20
Create a hook to let elements like iron-dropdown refine the behavior …

JanMiksovsky:
Submitting for review — please hold off on merging.

This is a small change to iron-overlay-behavior to enable a fix for PolymerElements/paper-dropdown-menu#141. This PR lets components using iron-overlay-behavior (like iron-dropdown) to refine the standard behavior of restoring focus to a node.

@valdrinkoshi I'll submit the PR to iron-dropdown shortly. Once that's done, please take a look.

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -247,7 +247,7 @@ context. You should place this element as a child of `<body>` whenever possible.
       // Used by __onNextAnimationFrame to cancel any previous callback.
       this.__raf = null;
       // Focused node before overlay gets opened. Can be restored on close.
-      this.__restoreFocusNode = null;
+      this._restoreFocusNode = null;
       this._ensureSetup();
     },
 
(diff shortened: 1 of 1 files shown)


valdrinkoshi:
I think it would be enough to just make `__restoreFocusNode` a protected variable `_restoreFocusNode`, without the `_restoreFocusToNode` method.
That way, the implementor of the behavior would have to just set it to `null` in order to avoid focus to be restored, e.g. `iron-dropdown` would do something like this:
```javascript
// Overridden from Polymer.IronOverlayBehaviorImpl
_onCaptureFocus: function(event) {
  // Focus went outside the dropdown
  if (!this.contains(event.target)) {
     // Avoid restoring focus after is closed.
     this._restoreFocusNode = null;
     // This will trigger `iron-overlay-canceled` and give
     // a chance to the user to prevent the canceling if necessary
     this.cancel(event);
  } else {
     Polymer.IronOverlayBehaviorImpl._onCaptureFocus.apply(this, arguments);
  }
}
```

JanMiksovsky:
Good idea, done.

valdrinkoshi:
Re scroll actions: #234 addresses it.
Re expose `__restoreFocusNode`: focus behavior can be controlled by users via `autofocus` attribute and `noAutoFocus, restoreFocusOnClose` properties.

Closing this for now

PR #229, opened 2017-01-31
useCapture for tap events

### LINKED ISSUE #225: manager should have a capturing event for tap
This change https://github.com/PolymerElements/iron-overlay-behavior/commit/d1adb7635f6a568afd0554e2588a3fc3b59ed092#diff-3a555b03d77a3dc28aad3488eed08943R43 made the listener non-capturing, allowing other listeners of the `tap` event to effectively prevent the closing of an overlay by calling `event.stopPropagation()`.
E.g. see this example https://jsbin.com/zewuqe/3/edit?html,output
The menu should close no matter which button is clicked outside it, instead `event.stopPropagation()` doesn't allow it.

@Caffeinix FYI

valdrinkoshi:
Fixes #225 by setting `useCapture = true` to the `tap` listener.

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -40,8 +40,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     this._backdropElement = null;
 
     // Enable document-wide tap recognizer.
-    Polymer.Gestures.add(document, 'tap', this._onCaptureClick.bind(this));
-
+    // NOTE: Use useCapture=true to avoid accidentally prevention of the closing
+    // of an overlay via event.stopPropagation(). The only way to prevent
+    // closing of an overlay should be through its APIs.
+    Polymer.Gestures.add(document, 'tap', null);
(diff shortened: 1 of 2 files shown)


joshcartme:
Hey @valdrinkoshi we had code that relied on using `event.stopPropogation()` to prevent an overlay from closing in some cases.  We do want to stay current on releases, could you suggest an alternative approach?

Thanks!

valdrinkoshi:
Hi @joshcartme, overlays will fire an `iron-overlay-canceled` event which can be prevented in order to avoid "canceling" the overlay.
From the [README.md](https://github.com/PolymerElements/iron-overlay-behavior#closing-and-canceling):
> When the dialog is canceled, the overlay fires an 'iron-overlay-canceled' event. Call `preventDefault` on this event to prevent the overlay from closing.

Hope it helps, let me know if you need some examples :)

joshcartme:
@valdrinkoshi I should have read more carefully. That should be easy enough to switch to, thanks for pointing me in the right direction!

PR #230, opened 2017-01-31
ensure to set styles when closing overlay gets detached before completing animations

### LINKED ISSUE #223: After reattaching, opened state is set to false but element remains visible
### Description
I'm using a paper-menu-button with closeOnActivate on. One of the options of the menu makes the container element to be removed from the DOM and replaced with a second element. Yet, from the second element I can choose to switch back to my first element withe menu. This brings the dropdown element into an inconsistent state where opened is false, but the element is visible.

### Expected outcome
When I reattach my first element back into the DOM, the menu should not be expanded.

### Actual outcome
When I come back to the first element, the menu is always already opened. After some deep investigation I found that the drop-down(iron-overlay) is closed (opened=false), but the style.hidden is set to '' (visible).
Clicking outside of the menu or selecting any menu option don't close the menu anymore because the dropdown state is already closed and the event doesn't propagate all the way to the end.
I realized that in the detached function, the call that was meant to happen via requestAnimationFrame() is cancelled.

FYI, for the moment I'm forcing a call to __openedChanged() when I detect this inconsistent state in the dropdown.

### LINKED ISSUE #247: Restore focus not working as expected
### Description
When restore-focus-on-close is set on a paper-dialog (which implements this behavior), the restoreFocusNode does not get its focus back, when the dialog is canceled and/or closed. The isAttached check added prevents the right part being called.

### Expected outcome
The focus is restored on the restoreFocusNode

### Actual outcome
The focus is lost and falls back to the first focusable item on the page.

### Steps to reproduce

1. Put a `paper-dialog` element with `restore-focus-on-close` attribute in the page.
2. Open the page in a web browser.
3. Click the `open-dialog` button.
4. Close the dialog.
5. The restore focus is not set on the button.

### Browsers Affected

- [x] Chrome

valdrinkoshi:
Fixes #223 by ensuring rendering tasks are completed if an overlay is detached before the animation is done.
Fixes #247 by applying focus if overlay was animating when detached.


iron-overlay-behavior.html
```diff
@@ -187,6 +187,16 @@
         window.cancelAnimationFrame(this.__raf);
         this.__raf = null;
       }
+      if (this.__isAnimating) {
+        this.__isAnimating = false;
+        // We got detached while animating, ensure styles are ok
+        // in case we get ever attached again.
+        // NOTE: it's too late to fire iron-overlay-closed as we're already detached.
```
valdrinkoshi: Yes, we could fire the event (e.g. use the host as the node from which to fire the event).
But I don't think it would be semantically correct to notify "the overlay was closed" when its animation is not done yet...similar to the scenario where an animating element that gets detached before finishing the animation wouldn't fire a "transitionend" event. WDYT?

iron-overlay-behavior.html
```diff
@@ -187,6 +187,16 @@
...
+        if (!this.opened) {
+          this.style.display = 'none';
+          this.style.zIndex = '';
```
valdrinkoshi: Good idea, will do 👌

valdrinkoshi:
@cdata ended up firing opened/closed event and updating styles on detached.

iron-overlay-behavior.html
```diff
@@ -188,6 +188,18 @@
         this.__raf = null;
       }
       this._manager.removeOverlay(this);
+
+      // We got detached while animating, ensure we show/hide the overlay
+      // and fire iron-overlay-opened/closed event!
+      if (this.__isAnimating) {
+        if (this.opened) {
+          this._finishRenderOpened();
```
e111077: I'm honestly impressed how you have a function for every possible situation

PR #233, opened 2017-02-21
fix global tap listener

### LINKED ISSUE #232: Polymer 2.0 - Polymer.Gestures.add has been privatized
This [Polymer 2.0 PR](https://github.com/Polymer/polymer/pull/4312) appears to have refactored the `Polymer.Gestures.add` method to `Polymer.Gestures._add` [here](https://github.com/Polymer/polymer/pull/4312/files#diff-20794b26841d6321df19e97b3f198111L342)

This affects [iron-overlay-manager#L43](https://github.com/PolymerElements/iron-overlay-behavior/blob/2.0-preview/iron-overlay-manager.html#L43) which now throws the following error:
image

TomK:
closes #232

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -40,7 +40,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     this._backdropElement = null;
 
     // Enable document-wide tap recognizer.
-    Polymer.Gestures.add(document, 'tap', this._onCaptureClick.bind(this));
+    Polymer.Gestures.addListener(document, 'tap', this._onCaptureClick.bind(this));
 
     document.addEventListener('focus', this._onCaptureFocus.bind(this), true);
     document.addEventListener('keydown', this._onCaptureKeyDown.bind(this), true);
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -198,7 +198,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           runAfterOpen(overlay, function() {
             var spy = sinon.spy(overlay, 'refit');
             overlay.fire('iron-resize');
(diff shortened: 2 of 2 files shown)


valdrinkoshi:
@TomK these should be addressed by polymer core and not elements (see https://github.com/Polymer/polymer/issues/3936 and https://github.com/Polymer/polymer/issues/4104).

TomK:
Thanks @valdrinkoshi
I saw those new PRs a little while ago. So I'll close this one now :)

TomK:
@valdrinkoshi i'll reopen this and revert the afterNextRender test changes.
I believe the global listener change is still valid as it is marked as deprecated.
Please advise.

valdrinkoshi:
Even if deprecated, we should use it in the polymer 1.x scenario. This behavior should work with polymer 2.x or 1.x.
In polymer 2.0-previe invoking `Polymer.Gestures.add` doesn't throw any warning or log, so I'd suggest to leave it as is. I'd just put a comment that `Polymer.Gestures.add` is deprecated in 2.x, WDYT?

TomK:
For this element, master/1.x is correct to leave as `Gestures.add` I would think.
For the 2.0 branch/version of this element I suppose it depends on process. If `add` can be removed at any time it might be safer to merge, preempting it's eventual removal.

The 2.0-preview branches of the polymer elements aren't intended to be used as hybrid are they? If they are, it can't be merged.

valdrinkoshi:
Yes, polymer 2.0 won't be a breaking change, and 2.0 elements will be hybrid and supposed to work with 1.x or 2.x 👌

TomK:
👍  this definitely can't be merged then because polymer v1.x does not include `Gestures.addListener`
https://github.com/Polymer/polymer/blob/37eef0e1934cf721bb79ded7e6493d9abd315ad7/src/standard/gestures.html

Closing this PR, thanks for your input!

PR #234, opened 2017-02-27
Block scrolling

### LINKED ISSUE #140: Overlays should not allow scrolling background content (esp in modal mode)
1. On https://elements.polymer-project.org/elements/paper-dialog?view=demo:demo/index.html&active=paper-dialog, resize the window to be very small
2. Notice: the background content is scrollable

Expectation:
The background content (esp with `with-backdrop`) is not scrollable. The paper-dialog should act like a proper modal.

cc @valdrinkoshi @cdata

valdrinkoshi:
Fixes #140, 
https://github.com/PolymerElements/iron-dropdown/issues/127,
https://github.com/PolymerElements/iron-dropdown/issues/126, 
https://github.com/PolymerElements/iron-dropdown/issues/120, 
https://github.com/PolymerElements/iron-dropdown/issues/112, 
https://github.com/PolymerElements/iron-dropdown/issues/101

## IronScrollManager 

Moved the `Polymer.IronDropdownScrollManager` into this repo, in order for all overlays to benefit from it. Now an overlay can prevent scrolling from happening outside it by setting the property `scrollAction = "lock"`.

## `scrollAction` property

Introduced this new property that allows to either:
- lock the scrolling from happening outside an opened overlay - scrollAction = "lock"
- cancel (aka close) an overlay on scroll outside - scrollAction = "cancel"
- reposition an overlay on scroll outside - scrollAction = "refit"

Here the PR which removes `Polymer.IronDropdownScrollManager` from iron-dropdown https://github.com/PolymerElements/iron-dropdown/pull/125


LarsDenBakker:
These are quite some serious bugs. Would be great if they get fixed by his. What's the status of this?

valdrinkoshi:
@cdata @e111077 mind taking a look at this? This PR moves the scroll manager from iron-dropdown into iron-overlay-behavior

iron-scroll-manager.html
```diff
@@ -0,0 +1,381 @@
...
+      'touchmove'
+    ];
+
+    /**
+     * The IronScrollManager is intended to provide a central source
+     * of authority and control over which elements in a document are currently
+     * allowed to scroll.
+     */
+
+    Polymer.IronScrollManager = {
```
e111077: I am assuming this is a direct copy-paste of `iron-dropdown-scroll-manager` with just name refactors. Is this correct?

iron-overlay-behavior.html
```diff
@@ -549,23 +565,98 @@
...
+
+    /**
+     * @param {boolean} isAttached
+     * @param {boolean} opened
+     * @param {boolean} withBackdrop
+     * @param {boolean} allowOutsideScroll
+     * @private
+     */
+    __lockOutsideScroll: function(isAttached, opened, withBackdrop, allowOutsideScroll) {
+      if (isAttached && opened && (withBackdrop || !allowOutsideScroll)) {
```
valdrinkoshi: Overlays with backdrop are expected to block scrolling (as requested in https://github.com/PolymerElements/iron-overlay-behavior/issues/140)

iron-scroll-manager.html
```diff
@@ -0,0 +1,381 @@
...
+        if ('deltaX' in event) {
+          // do nothing, values are already good.
+        }
+        // Safari has scroll info in `wheelDeltaX/Y`.
+        else if ('wheelDeltaX' in event && 'wheelDeltaY' in event) {
+          info.deltaX = -event.wheelDeltaX;
+          info.deltaY = -event.wheelDeltaY;
+        }
+        // IE10 has only vertical scroll info in `wheelDelta`.
+        else if ('wheelDelta' in event) {
```
valdrinkoshi: This is an addition compared to `IronDropdownScrollManager`

iron-scroll-manager.html
valdrinkoshi: Yes, modulo an addition that I pointed in line 360 https://github.com/PolymerElements/iron-overlay-behavior/pull/234/files#diff-893924a58950c7613a35f77547102b7eR360

iron-overlay-behavior.html
```diff
@@ -549,23 +565,98 @@
...
+      }
+    },
+
+    /**
+     * @param {boolean} isAttached
+     * @param {boolean} opened
+     * @private
+     */
+    __updateScrollListener: function(isAttached, opened) {
+      if (isAttached && opened) {
```
valdrinkoshi: Yes, it saves the scroll position (causing a repaint) also when that might not be needed :/
This must be rethought, maybe having a multi-property observer..? I'm kinda scared of all these properties being observed tbh :x

iron-overlay-behavior.html
e111077: Though that's a feature that should be implemented; I'm really scared about the performance hit this would introduce to such a common use case. @cdata do you have any input on this?

iron-overlay-behavior.html
e111077: Can't you just add `allowOutsideScroll` (and depending on the previous comment `withBackdrop`) as another listened variable? I don't think it would add too much overhead.

valdrinkoshi:
@e111077 feedback incorporated, & changed the implementation to use `scrollAction` instead. Updated PolymerElements/iron-dropdown#125 as well

iron-overlay-behavior.html
```diff
@@ -549,23 +570,113 @@
...
+
+    /**
+     * @param {string=} scrollAction
+     * @return {boolean}
      * @private
      */
-    __onNextAnimationFrame: function(callback) {
-      if (this.__raf) {
-        window.cancelAnimationFrame(this.__raf);
+    __isValidScrollAction: function(scrollAction) {
```
e111077: Do you see any benefit to make this protected / overridable as well as `__updateScollObservers?`

valdrinkoshi: I was thinking to have a possible value of `scrollAction` to be `custom` or directly be the name of the callback to be invoked, but I refrained from doing it in this PR for now.
I'd like to see if the 3 actions provided already cover most of the cases and avoid exposing features that haven't been requested.

e111077: Sounds good. Anyway it's easier to add a feature than remove.

e111077:
@valdrinkoshi This is a really big PR that introduces a huge change. I'd still like to see a second reviewer give it a scan

valdrinkoshi:
@e111077 agreed 👌 hoping @cdata will come to the rescue! 👀

iron-overlay-behavior.html
```diff
@@ -265,8 +286,8 @@
 
       this.__isAnimating = true;
 
-      // Use requestAnimationFrame for non-blocking rendering.
-      this.__onNextAnimationFrame(this.__openedChanged);
+      // Deraf for non-blocking rendering.
+      this.__deraf('__openedChanged', this.__openedChanged);
```
cdata: 👍

iron-overlay-behavior.html
```diff
@@ -183,10 +202,12 @@
     detached: function() {
       Polymer.dom(this).unobserveNodes(this._observer);
       this._observer = null;
-      if (this.__raf) {
-        window.cancelAnimationFrame(this.__raf);
-        this.__raf = null;
+      for (var cb in this.__rafs) {
+        if (this.__rafs[cb]) {
```
cdata: I have a personal style preference for checking more explicitly in cases like this e.g., `this.__rafs[cb] != null` or whatever is most appropriate for the condition.

iron-overlay-behavior.html
```diff
@@ -549,23 +570,113 @@
     },
 
     /**
-     * Executes a callback on the next animation frame, overriding any previous
-     * callback awaiting for the next animation frame. e.g.
-     * `__onNextAnimationFrame(callback1) && __onNextAnimationFrame(callback2)`;
-     * `callback1` will never be invoked.
-     * @param {!Function} callback Its `this` parameter is the overlay itself.
+     * Delays the execution of a callback to the next animation frame.
+     * To cancel a previous job, pass callback = undefined
```
cdata: Using an undefined second argument the cancel a previous job is subtle and impossible to intuit without reading this comment. Consider factoring raf cancellation into an explicit method.

iron-overlay-behavior.html
```diff
@@ -549,23 +570,113 @@
...
+     * @param {!string} jobname
+     * @param {Function=} callback Always bound to `this`
+     * @private
+     */
+    __deraf: function(jobname, callback) {
+      var rafs = this.__rafs;
+      if (rafs[jobname]) {
```
cdata: Similar comment to the one above, re: explicit checking.

iron-overlay-behavior.html
valdrinkoshi: You're right. Actually, this "feature" is not used anywhere, so I'll just expect it to be always defined 👌

Issue #235, opened 2017-03-24
TypeError: t.getBoundingClientRect is not a function


camargo:
### Description
This is the same issue as #44.

### Expected outcome

No error should be thrown.

### Actual outcome

Getting the error in the console:

`TypeError: t.getBoundingClientRect is not a function`

This is being triggered by https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L46

![capture](https://cloud.githubusercontent.com/assets/683355/24310335/bdb98906-108c-11e7-8e9b-ed19ece13b6a.PNG)

### Steps to reproduce

1. Open Firefox
2. Go to https://www.webcomponents.org/element/PolymerElements/iron-overlay-behavior/demo/demo/index.html
3. Make your window smaller than your screen
4. Click down on the `Plain overlay` button
5. Drag your cursor **while still clicked down** outside of the browser window
6. Release click outside the browser window
7. The error will be in the console

### Browsers Affected

- [ ] Chrome
- [X] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

valdrinkoshi:
In Firefox apparently the `event.target` is `document`, while in the other browsers is `document.documentElement` (aka `<html>`), see a simpler repro http://jsbin.com/mabisob/2/edit?html,console,output
This issue should be resolved in `Polymer.Gestures`, will move the issue there.

valdrinkoshi:
This issue was moved to Polymer/polymer#4459

camargo:
@valdrinkoshi Right now the gesture is being added to `document` (https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L46). Could we add it to `body` or `html`. That could fix the issue.

valdrinkoshi:
I'd do `document.documentElement` for all of these https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L46-L49 👌
PRs welcome 😃

camargo:
@valdrinkoshi Alright that should do it, please ignore the first one

PR #239, opened 2017-03-29
Update iron-overlay-manager.html with document.documentElement for taps

camargo:
use document.documentElement for Gestures and tap.
fixes https://github.com/PolymerElements/iron-overlay-behavior/issues/235

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -43,8 +43,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // NOTE: Use useCapture=true to avoid accidentally prevention of the closing
     // of an overlay via event.stopPropagation(). The only way to prevent
     // closing of an overlay should be through its APIs.
+    // NOTE: enable tap on <html> to workaround Polymer/polymer#4459
     Polymer.Gestures.add(document.documentElement, 'tap', null);
-    document.documentElement.addEventListener('tap', this._onCaptureClick.bind(this), true);
+    document.addEventListener('tap', this._onCaptureClick.bind(this), true);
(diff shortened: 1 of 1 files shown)


camargo:
@valdrinkoshi Is this better?

iron-overlay-manager.html
```diff
@@ -43,8 +43,8 @@
     // NOTE: Use useCapture=true to avoid accidentally prevention of the closing
     // of an overlay via event.stopPropagation(). The only way to prevent
     // closing of an overlay should be through its APIs.
-    Polymer.Gestures.add(document, 'tap', null);
-    document.addEventListener('tap', this._onCaptureClick.bind(this), true);
+    Polymer.Gestures.add(document.documentElement, 'tap', null);
```
valdrinkoshi: This is the only line that needs to change, as we just need to enable the Polymer Gestures to recognize `tap`. It would be great to add also a comment on why we do this & point to the polymer issue, e.g.
> // NOTE: enable tap on `<html>` to workaround Polymer/polymer#4459

I just want to avoid any possible regressions, e.g. someone might be listening for the same event with useCapture = true and stop the propagation:
```js
document.addEventListener('tap', function(e) {
  e.stopPropagation();
}, true)
```
If we listen for `tap` on `document.documentElement`, we wouldn't capture it.

valdrinkoshi: Here an example showing the possible regression we might introduce by changing line 47
http://jsbin.com/qubifad/1/edit?html,console,output

valdrinkoshi:
Yay!! Build is 💚 ! Shipping this, thanks @camargo again for the contribution!

valdrinkoshi:
shipped in `v1.10.4` 🎉 :shipit:

PR #246, opened 2017-05-29
Bugfix for issue #247

### LINKED ISSUE #195: Loses focus on emulated device mode and mobile devices
### Description
When `restore-focus-on-close` is set, the `restoreFocusNode` loses focus when overlay closes on emulated device mode on Chrome and mobile devices. However, on desktop, `restoreFocusNode` focuses and can be seen with focus ring when overlay closes.

### Expected outcome
`restoreFocusNode` focuses with visible focus ring.

### Actual outcome
`restoreFocusNode` loses focus when overlay closes.

### Live Demo
http://jsbin.com/popukoruye/edit?html,output

### Browsers Affected
- [x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

### LINKED ISSUE #247: Restore focus not working as expected

ruud:
closes #247

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -259,7 +259,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       // Defer any animation-related code on attached
       // (_openedChanged gets called again on attached).
-      if (!this.isAttached) {
+      if (this.isAttached === undefined) {
         return;
       }
 


ruud:
@e111077 , could you also have a look at this focus related issue? Thanks in advance

ruud:
This is only an issue when the dialog is detached from the DOM before the animation cycle is complete.

iron-overlay-behavior.html
```diff
@@ -259,7 +259,7 @@
 
       // Defer any animation-related code on attached
       // (_openedChanged gets called again on attached).
-      if (!this.isAttached) {
+      if (this.isAttached === undefined) {
```
valdrinkoshi: This would cause executing `__openedChanged` (and trigger relayouts) also when the overlay is detached.
The fix should be done in `detached`, where we need to check if the overlay `__isAnimating` and call `_applyFocus()` if so. Addressing this in https://github.com/PolymerElements/iron-overlay-behavior/pull/230

valdrinkoshi:
Closing in favor of #230

PR #252, opened 2017-10-03
backdrop z-index below top overlay with-backdrop

### LINKED ISSUE #124: Wrong backdrop z-index for multiple overlays

valdrinkoshi:
Fixes #124.
Given 2 overlays with backdrop, the last one to be opened should have the backdrop below it, meaning the backdrop should be rendered between the 2 overlays.

### DIFF
--- a/test/iron-overlay-behavior.html
+++ b/test/iron-overlay-behavior.html
@@ -1196,9 +1196,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             var style1Z = parseInt(window.getComputedStyle(overlay1).zIndex, 10);
             var style2Z = parseInt(window.getComputedStyle(overlay2).zIndex, 10);
             var bgStyleZ = parseInt(window.getComputedStyle(overlay1.backdropElement).zIndex, 10);
-            assert.isAbove(style2Z, style1Z, 'overlay1 above overlay2');
-            assert.isAbove(style2Z, bgStyleZ, 'backdrop above overlay2');
-            assert.isBelow(style1Z, bgStyleZ, 'backdrop below overlay1');
+            assert.isAbove(style2Z, style1Z, 'overlay2 above overlay1');
+            assert.isAbove(style2Z, bgStyleZ, 'overlay2 above backdrop');
+            assert.isBelow(style1Z, bgStyleZ, 'overlay1 below backdrop');
             done();
           });
         });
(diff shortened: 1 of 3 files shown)


valdrinkoshi:
I tried to explain it [here](https://github.com/PolymerElements/iron-overlay-behavior/issues/124#issuecomment-333925976), basically I consider the current behavior a bug, because if you were to open 1 overlay with-backdrop, then another overlay (no backdrop), the rendering would be the same as if you were to open 1 overlay with-backdrop followed by another one with-backdrop. In other words, the lastly opened overlay with-backdrop is not hiding the overlays below it. WDYT?

test/iron-overlay-behavior.html
```diff
@@ -1154,11 +1154,26 @@
       test('newest overlay appear on top', function(done) {
         runAfterOpen(overlay1, function() {
           runAfterOpen(overlay2, function() {
-            var styleZ = parseInt(window.getComputedStyle(overlay1).zIndex, 10);
-            var style1Z = parseInt(window.getComputedStyle(overlay2).zIndex, 10);
+            var style1Z = parseInt(window.getComputedStyle(overlay1).zIndex, 10);
+            var style2Z = parseInt(window.getComputedStyle(overlay2).zIndex, 10);
+            var bgStyleZ = parseInt(window.getComputedStyle(overlay1.backdropElement).zIndex, 10);
+            assert.isAbove(style2Z, style1Z, 'overlay1 above overlay2');
```
e111077: aren't all of these messages backwards? doesn't this mean overlay2 has a higher z index than overlay1 thus 2  is above 1?

valdrinkoshi: The way I was considering it is: the message should reflect the wrong state rather than the correct state, as in "Error! overlay1 above overlay2!". Lol, lemmy be positive & write this message to reflect the right state, as in "All good, overlay2 is above overlay1" :p

valdrinkoshi: @e111077 done in https://github.com/PolymerElements/iron-overlay-behavior/pull/252/commits/5050abeeb1a08160e44c80b4b64622f5482936df

PR #255, opened 2017-11-21
Capture scroll events

### LINKED ISSUE #254: find scrollable parent instead of defaulting to document.body
_From @valdrinkoshi on August 19, 2016 1:49_

Currently the `iron-dropdown` will assume only body is scrollable. But if scroll happens in another scrollable parent the flow is broken, e.g. `iron-dropdown` that allows outside scrolls doesn't get its position updated http://jsbin.com/jucoda/2/edit?html,output

_Copied from original issue: PolymerElements/iron-dropdown#107_

valdrinkoshi:
Fixes #254.
Listen for `scroll` events in all shadowRoots hosting the overlay, up to the main `document`.

Note that:
- if the scroll happens inside a shadowRoot that is not in the overlay's path, it won't be captured
- this doesn't guarantee preventing scrolls happening in any node different from `document`

e.g. see http://jsbin.com/wewemu/6/edit?html - scroll events happening in `scrollable_0` cannot be handled by the opened overlays, as they are contained in `scrollable_1`. For these cases, the user is expected to listen to the scroll events on different targets, and then calling `refit` or `cancel` on the overlay.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -285,6 +285,36 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       </simple-overlay>
     </template>
   </demo-snippet>
+  
+  <h3>Use <code>scroll-action</code> to lock, refit or cancel an overlay on scroll.</h3>
+  <demo-snippet>
+    <template>
+      <span>
+        <button onclick="lockOnScroll.open()">locks scroll</button>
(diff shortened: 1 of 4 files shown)


benoitjchevalier:
🎉

davidrleonard:
Very nice, thanks for making this fix. Solves a lot of issues for us.

PR #261, opened 2018-02-02
Add TypeScript declarations to iron-overlay-behavior.

aomarks:
This PR adds TypeScript declarations generated by https://github.com/Polymer/gen-typescript-declarations/

These declarations can be re-generated by running `npm run update-types`.

Tracker: https://github.com/Polymer/gen-typescript-declarations/issues/79


iron-overlay-behavior.d.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * The node being focused.
+     */
+    _focusedChild: Node|null;
+
+    /**
+     * The backdrop element.
+     */
+    backdropElement(): any;
```
valdrinkoshi: Shouldn't this be of type `Element`? 
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L157

iron-overlay-behavior.d.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Returns the node to give focus to.
+     */
+    _focusNode(): any;
```
valdrinkoshi: type `Node`
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L165

iron-overlay-behavior.d.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Array of nodes that can receive focus (overlay included), ordered by `tabindex`.
+     * This is used to retrieve which is the first and last focusable nodes in order
+     * to wrap the focus for overlays `with-backdrop`.
+     *
+     * If you know what is your content (specifically the first and last focusable children),
+     * you can override this method to return only `[firstFocusable, lastFocusable];`
+     */
+    _focusableNodes(): any;
```
valdrinkoshi: type `Array<Node>`
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L178

iron-overlay-behavior.d.ts
```diff
@@ -0,0 +1,263 @@
...
+   *   The element is styled to appear on top of other content by setting its `z-index` property. You
+   *   must ensure no element has a stacking context with a higher `z-index` than its parent stacking
+   *   context. You should place this element as a child of `<body>` whenever possible.
+   *
+   *   
+   */
+  interface IronOverlayBehavior {
+  }
+
+  const IronOverlayBehavior: object;
```
valdrinkoshi: how is it expressed that `IronOverlayBehavior` is composed by other behaviors?
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L796

iron-overlay-behavior.d.ts
```diff
@@ -0,0 +1,263 @@
...
+    /**
+     * Refits if the overlay is opened and not animating.
+     */
+    _onIronResize(): any;
+
+    /**
+     * Will call notifyResize if overlay is opened.
+     * Can be overridden in order to avoid multiple observers on the same node.
+     */
+    _onNodesChange(): any;
```
valdrinkoshi: is `_manager` property being omitted on purpose because it's private?
https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L132

iron-overlay-behavior.d.ts
aomarks: Huh, that's a good catch. We don't handle behavior arrays right now. Filed https://github.com/Polymer/gen-typescript-declarations/issues/80 and will update this PR when done.

iron-overlay-behavior.d.ts
aomarks: Done. Actually it should be `!Array<!Node>` or else it would be a nullable array of nullable nodes, since non-primitive types are nullable by default in Closure. Updated the other cases I found too.

iron-overlay-behavior.d.ts
aomarks: Done.

iron-overlay-behavior.d.ts
aomarks: Done. Looks like we're sometimes using the `@type` annotation for getters, but it should be `@return`. Updated. Also updated the other cases I found.

iron-overlay-behavior.d.ts
aomarks: Yes, we don't emit private things.

iron-scroll-manager.html
```diff
@@ -56,6 +56,7 @@
       /**
        * The current element that defines the DOM boundaries of the
        * scroll lock. This is always the most recently locking element.
+       * @return {!Node}
```
valdrinkoshi: `currentLockingElement` can be undefined - this was wrongly annotated previously as it was saying that `currentLockingElement` can be null

aomarks:
> LGTM, though it would be nice to not have to duplicate the escape codes and message for each element.

I'll update the tedium script to do this for all the repos (https://github.com/PolymerLabs/tedium/pull/63), although did you have something else in mind? I guess I could make the generator itself have a flag that errors if anything has changed?

iron-overlay-behavior.d.ts
aomarks: Updated to add the other behaviors. See interface above.

iron-scroll-manager.html
aomarks: Done.

iron-focusables-helper.d.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
aomarks: @valdrinkoshi Is `IronFocusablesHelper` supposed to be something users use, or is it fine that this file is empty? It doesn't have any annotations, so analyzer doesn't pick it up as an interesting feature. Same question for `IronScrollManager`.

valdrinkoshi: All managers/helpers are singletons available on `window` and expose public methods, so can be used by users: `IronFocusablesHelper, IronOverlayManager, IronScrollManager`
I'm sure users rely on the public methods of `IronFocusablesHelper` and `IronOverlayManager` as they're been around for a while.

aomarks: I've added some annotations to generate `IronScrollManager`, although it is missing a getter. `IronOverlayManager` is probably too difficult for Analyzer to understand at all right now. Filed https://github.com/PolymerElements/iron-overlay-behavior/issues/262 with details on both.

I'd like to merge this PR despite these missing types, because this element is a dependency of a lot of other ones. I'd like to go back and add some manual typings for the managers a bit later, if that's ok.

valdrinkoshi:
Sounds good, LGTM

PR #271, opened 2018-03-10
Prevent open again when click on the trigger!

AliMD:
In `iron-dropdown` element when user click on the trigger when the dropdown is open, dropdown close and after that it open again and use see a flash on dropdown-content

To prevent this we need to call `cancel` asynchronous to make a delay.

Before this PR I simply try to `if (this.opened) return;` in my trigger function but dropdown closed before trigger event then I forced to make this PR!

### DIFF
--- a/iron-overlay-behavior.html
+++ b/iron-overlay-behavior.html
@@ -462,9 +462,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @protected
      */
     _onCaptureClick: function(event) {
-      if (!this.noCancelOnOutsideClick) {
-        this.cancel(event);
-      }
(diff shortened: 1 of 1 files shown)


valdrinkoshi:
Hi @AliMD, you can listen for `iron-overlay-canceled` event, and invoke `event.preventDefault()` to prevent the overlay from closing, e.g. see how `<paper-menu-button>` does it here:
- add listener: https://github.com/PolymerElements/paper-menu-button/blob/99036edd407b885e55f9b4550c71679a3c088b4f/paper-menu-button.html#L132
- call preventDefault: https://github.com/PolymerElements/paper-menu-button/blob/99036edd407b885e55f9b4550c71679a3c088b4f/paper-menu-button.html#L466-L475

PR #279, opened 2018-04-13
Fix modulizer

e111077:
Ignore the million commits, I had to commit to run modulizer; I'm going to have to squash them.

## iron-scroll-manager.html
Added a declaration for `_boundScrollHandler` as modulizer does not play nice with namespaces and runs in strict mode.

Modulizer cannot work with getters in namespaces. It converts them to functions, so to keep backwards compatibility, I had to create a `getCurrentLockingElement ` function.

I had to wrap the `.bind(this)` in a try catch because modulizer would get rid of the `Polymer. IronScrollManager` namespace, but would replace the `this` in the `bind` with `Polymer. IronScrollManager` which would always fail. I could not get around this; it looks like modulizer would read the closure.

## test/iron-overlay-behavior.html
I added some `var`s for strict mode

## everything else
I added commas to the end of each property definition in the properties block so it looks nicer formatted with clang-format

### DIFF
--- a/iron-scroll-manager.html
+++ b/iron-scroll-manager.html
@@ -43,6 +43,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   ];
   // must be defined for modulizer
   var _boundScrollHandler;
+  if (!window.Polymer) {
+    window.Polymer = {};
+  }
 
   /**
    * The IronScrollManager is intended to provide a central source
--- a/iron-scroll-manager.html
+++ b/iron-scroll-manager.html
@@ -101,8 +101,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return false;
       }
 
-      scrollLocked = !!lockingElement &&
-          lockingElement !== element &&
+      scrollLocked = !!lockingElement && lockingElement !== element &&
           !this._composedTreeContains(lockingElement, element);
 
       if (scrollLocked) {
@@ -240,8 +239,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     _lockScrollInteractions: function() {
       try {
-        this._boundScrollHandler =
-            this._boundScrollHandler || this._scrollInteractionHandler.bind(scope);
+        this._boundScrollHandler = this._boundScrollHandler ||
+            this._scrollInteractionHandler.bind(scope);
       } catch (e) {
         this._boundScrollHandler =
             this._boundScrollHandler || this._scrollInteractionHandler;
--- a/iron-scroll-manager.html
+++ b/iron-scroll-manager.html
@@ -228,14 +228,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _boundScrollHandler: undefined,
 
     _lockScrollInteractions: function() {
-      // workaround for modulizer not handling bind properly
-      if (!this._boundScrollHandler) {
-        try {
-          this._boundScrollHandler = this._scrollInteractionHandler.bind(this);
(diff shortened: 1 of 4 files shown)


iron-scroll-manager.html
```diff
@@ -62,6 +64,17 @@
...
     },
 
+    /**
+     * The current element that defines the DOM boundaries of the
+     * scroll lock. This is always the most recently locking element.
+     * Required for modulizer
+     *
+     * @return {!Node|undefined}
+     */
+    getCurrentLockingElement: function() {
```
valdrinkoshi: instead of adding this, we could convert `currentLockingElement` to be just a property (not a getter)
```js
    /**
      * The current element that defines the DOM boundaries of the
      * scroll lock. This is always the most recently locking element.
      *
      * @type {!Node|undefined}
      */
     currentLockingElement: undefined,
```

and update it in `pushScrollLock ` and `removeScrollLock`
```js
this.currentLockingElement = this._lockingElements[this._lockingElements.length - 1];
```

iron-scroll-manager.html
```diff
@@ -224,8 +237,14 @@
     _boundScrollHandler: undefined,
 
     _lockScrollInteractions: function() {
-      this._boundScrollHandler =
-          this._boundScrollHandler || this._scrollInteractionHandler.bind(this);
+      // workaround for modulizer not handling bind properly
+      if (!this._boundScrollHandler) {
```
valdrinkoshi: This looks quite ugly :x
two options to explore:
1. Use only `_scrollInteractionHandler` and set at the bottom of the file 
```js
Polymer.IronScrollManager._scrollInteractionHandler = 
  Polymer.IronScrollManager._scrollInteractionHandler.bind(Polymer.IronScrollManager);
```
2. avoid using `.bind()` and rename `_scrollInteractionHandler` to `handleEvent`, so that we can add/remove listeners to `this` - https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

e111077: 1 will not work with modulizer because modulizer removes the entire `Polymer.IronScrollManager` definition and will keep it inside that bind (thus throwing Polymer is undefined errors)

2, I don't think I fully understand your explanation, but the issue is that modulizer would remove the entire `Polymer.IronScrollManager` namespace and simply export the functions. Since, it removes the namespace and it's just naked functions in a module, then `this` is undefined.

I think a good 3 would be to make this into a class instead of a namespace, but I believe that will break IE in Polymer 1.

valdrinkoshi: you can add an event listener like this
```js
document.addEventListener('scroll', this);
```
^ when scroll is fired, `this.handleEvent(event)` is invoked.
So the suggestion is to completely remove `_boundScrollHandler` and rename `_scrollInteractionHandler` to `handleEvent`

e111077:
I reverted the try catch. I was given the 👍 to fix this after modulization, but I still need the other fixes in there (delclaring variables)

e111077:
also we need to give fixing this element priority as it's blocking Kamino

e111077:
Looks like we're going to fix it in modulizer and when in namespace that uses `this` in a function to transform it to `undefined` instead. For now we're gonna have to do it manually

e111077:
Travis might be test thrashing still

e111077:
we will have to just manually remove the bind when we run modulizer pass so it's more of a modulizer issue than a `iron-overlay-behavior`.

relevant issue is Polymer/polymer-modulizer#345

e111077:
Tests pass locally

PR #289, opened 2018-10-23
Make IronFocusablesHelper a singleton

rictic:
This is easier for closure compiler to read and optimize correctly, and is otherwise a no-op.

Similar to the change we made to the other helper in this repo.

Internalized as cl/218204453


rictic:
Hm, what's your read of this weird test output?

e111077:
never seen those before. I'm rerunning

rictic:
Thanks!

PR #291, opened 2019-03-04
Use @package visibility for private exports.

rictic:
`@private` doesn't make sense as a visibility for an exported binding, and jscompiler will begin warning about this soon.

Downstreamed as cl/236669868 and cl/236397214


rictic:
Tests are failing, but not from this change. Something changed in dependencies I think, as rerunning the previously passing master commit on travis also fails: https://travis-ci.org/PolymerElements/iron-overlay-behavior/builds/444928303

PR #292, opened 2019-03-20
Fixed _collectTabbableNodes to use shadowRoot as well

yosilevy:


aigan:
This should be merged

PR #294, opened 2019-03-20
Collect tabbable nodes fix

yosilevy:


yosilevy:
Tests are failing without changing anything...

PR #295, opened 2019-05-01
Minor Closure compilation fix

aomarks:
Upstreaming cl/245921774


aomarks:
Travis is already failing at master with the same error

PR #299, opened 2019-08-26
Update deprecated function call

mgiuffrida:
Replace call to deprecated Polymer.Gestures.add function with Polymer.Gestures.addListener.

Fixes a Closure compiler error (JSC_DEPRECATED_PROP_REASON).

### DIFF
--- a/iron-overlay-manager.html
+++ b/iron-overlay-manager.html
@@ -44,7 +44,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // of an overlay via event.stopPropagation(). The only way to prevent
     // closing of an overlay should be through its APIs.
     // NOTE: enable tap on <html> to workaround Polymer/polymer#4459
-    Polymer.Gestures.add(document.documentElement, 'tap', null);
+    Polymer.Gestures.addListener(document.documentElement, 'tap', null);
     document.addEventListener('tap', this._onCaptureClick.bind(this), true);
     document.addEventListener('focus', this._onCaptureFocus.bind(this), true);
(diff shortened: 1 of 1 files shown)


mgiuffrida:
@aomarks can you PTAL? We have had to patch this into Chromium (crbug.com/997852).

aomarks:
Looks correct (https://polymer-library.polymer-project.org/3.0/docs/devguide/gesture-events). Thanks for the PR.

aomarks:
Oops, sorry. I just realized that this PR was targeting the `1.x` branch. I'm going to need to roll this back. Polymer 1.x only defines `Gestures.add`: https://github.com/Polymer/polymer/blob/1.x/src/standard/gestures.html#L341, so this change doesn't make sense.

Polymer 2 and 3 do define `add` as a `@deprecated` alias for `addListener`:
2.x: https://github.com/Polymer/polymer/blob/2.x/lib/utils/gestures.html#L1104
3.x: https://github.com/Polymer/polymer/blob/master/lib/utils/gestures.js#L1081

And the `2.x` and `3.x` versions of this library do indeed call `.add` (https://github.com/PolymerElements/iron-overlay-behavior/blob/2.x/iron-overlay-manager.html#L48 and https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.js#L49).

Are you sure you're using the right version of Polymer and/or this element library? If you'd still like to fix it in the `2.x` and/or latest branches, please send another PR.

mgiuffrida:
I see. Chromium is pulling down Polymer 2.x so that's the branch I need to target -- I'll open a PR for master first though.

PR #300, opened 2019-09-23
Update deprecated function call

mgiuffrida:
Replace call to deprecated Polymer.Gestures.add function with Polymer.Gestures.addListener.

Fixes a Closure compiler error (JSC_DEPRECATED_PROP_REASON).


mgiuffrida:
@aomarks PTAL at this version to land the change directly on master (which I should have done in the first place).

aomarks:
Thanks Michael!

PR #304, opened 2019-11-26
Fix flaky tests.

bicknellr:


bicknellr:
The 'try again if focusing fails' thing is a pretty ugly hack so I'm going to run this through internal tests to make sure it doesn't cause some major issue.

bicknellr:
Yup, caused an issue w.r.t. focus delegation. Details here: 81e3720

PR #305, opened 2019-11-26
Restore focus if the node with focus in the overlay is distributed into it.

bicknellr:
Before this PR, focus was only restored when closing an overlay if there was no focused element (or `<body>`) or the focused element was a shadow-including descendant of the overlay. This prevented focus from being restored if the focused element inside the overlay was distributed into the overlay through a slot:

```html
<outer-element>
  #shadow-root
    <element-with-iron-overlay-behavior>
      <slot name="in-overlay">
    </element-with-iron-overlay-behavior>
  #/shadow-root
  <input slot="in-overlay">
</outer-element>
```

This PR changes the test to check that the focused node is any descendant of the overlay in the *flattened* tree.

### DIFF
--- a/iron-overlay-behavior.js
+++ b/iron-overlay-behavior.js
@@ -763,7 +763,8 @@ export const IronOverlayBehaviorImpl = {
 
 };
 
-const composedParent = node => node.assignedSlot || node.parentNode || node.host;
+const composedParent = node =>
+    node.assignedSlot || node.parentNode || node.host;
 
 const composedContains = (ancestor, descendant) => {
   for (let element = descendant; element; element = composedParent(element)) {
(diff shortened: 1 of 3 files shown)


iron-overlay-behavior.js
```diff
@@ -763,6 +763,18 @@ export const IronOverlayBehaviorImpl = {
 
 };
 
+const composedParent = node =>
+    node.assignedSlot || node.parentNode || node.host;
```
e111077: I'm on the fence of this is scary and this is brilliant, but it seems to terminate in 100% of the cases I can think of. I wonder if you can achieve this with event listeners...

e111077: Actually no, it won't work with listeners in the case of

```html
<a-a>
  #root1
    <my-dialog>
      #root2
        <ancestor @listener=${thisWontBeCalled}>
          <slot></slot>
        </ancestor>
      #/root2

      <outerAncestor @listener=${thisWillBeCalled}>
        <slot></slot>
      </outerAncestor>
    </my-dialog>
  #/root1

  <div id="target"></div>
</a-a>
```

e111077: example:

https://stackblitz.com/edit/typescript-litelement-example-fr9vvp?file=Cat.ts

PR #309, opened 2021-05-05
feat: add [iron-focusable] to mark as focusable

### LINKED ISSUE #282: Focus trapping logic doesn't work with non-Polymer elements
### Description
Focus traping logic is not compatible with elements that are not extending **PolymerElement**. Moreover, it won't work properly even if at least one of parent shadow roots is not a Polymer element.

As far as I can tell the main reason for that is because of the way [deepActiveElement](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L82) getter of the `iron-overlay-manager` is implemented. It expects that the shadow root is accessible via `root` property.
**LitElement**, for example, contains the shadow root in the `_root` property, so the `deepActiveElement` won't get a proper active element for it.

Suggestion: why not use [shadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/Element/shadowRoot) to access the shadow root instead?

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

e111077:
addresses but not f1x #282 and PolymerElements/paper-dialog#125

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ options.
 
 In addition, `with-backdrop` will wrap the focus within the content in the light DOM.
 Override the [`_focusableNodes` getter](#Polymer.IronOverlayBehavior:property-_focusableNodes)
-to achieve a different behavior.
+to achieve a different behavior. Additionally, you can add the `iron-focusable` attribute to
+any element in the light DOM to mark it as focusable.
 
 ### Limitations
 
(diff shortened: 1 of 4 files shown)


bicknellr:
Would this issue also be fixed by #292? (which I've neglected to review forever...)

rictic (approved):

e111077:
That would indeed also work

e111077:
Also I think we'd need both in case the host has a custom `.focus` rather than the `focus-within` inside

